PR #117, opened 2014-07-23
Superflux

MartinHN:
still non fully working superflux extractor :

superflux

bug seems to come from the output of superfluxPeaks algorithm if not in raw_mode


dbogdanov:
I've cherry-picked commits to add stuff related to SuperFlux. What is exactly "not fully working"?
See some TODO comments in essentia master in MaxFilter and TriangleBands algorithms.

PR #127, opened 2014-08-21
Now guards against positions or amplitudes vectors being empty, raising ...

leighsmith:
...an exception, to avoid segfaults


dbogdanov:
This check for empty peak vectors is already implemented in a different way.

PR #189, opened 2014-11-21
Add FreeBSD support

tungol:
This change enables essentia and examples to compile on FreeBSD. Besides roguevector.h, the other sticking point is atomic.h, which compiles fine if you pass in -std=c++11 and bypass the OS specific implementations.


dbogdanov:
Thank you! We require some steps to be able to merge contributions:
https://github.com/MTG/essentia#how-to-contribute

Issue #122, opened 2014-08-08
Error compiling with configuration --with-cpptests: 'DSO missing from command line'


aberrian:
Hi,

I'm a total beginner here, attempting to compile Essentia for the first time.  I've installed all the dependencies, and I've configured before building.  Then, after I run ./waf, I get the following error at the end of the build:

```
/usr/bin/ld: test/3rdparty/gtest-1.6.0/src/gtest-all.cc.1.o: undefined reference to symbol 'pthread_key_delete@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

This yields the following build failure message:
```
Build failed
 -> task in 'basetest' failed (exit status 1): 
	{task 139780899558160: cxxprogram gtest-all.cc.1.o,test_audioloader.cpp.1.o,test_composite.cpp.1.o,test_connectors.cpp.1.o,test_copy.cpp.1.o,test_fileoutput.cpp.1.o,test_main.cpp.1.o,test_network.cpp.1.o,test_networkparser.cpp.1.o,test_parameter.cpp.1.o,test_peak.cpp.1.o,test_pool.cpp.1.o,test_scheduler.cpp.1.o,test_stringutil.cpp.1.o,test_treetraversal.cpp.1.o,test_vectorinput.cpp.1.o,test_vectoroutput.cpp.1.o -> basetest}
[...]

PR #203, opened 2014-12-19
Fix compiling error

BigCrunsh:
see https://github.com/MTG/essentia/issues/122


dbogdanov:
Thanks! We require additional steps to be able to merge contributions: http://essentia.upf.edu/contribute

PR #217, opened 2015-01-21
Experimental windows support

carthach:
Hey,

Here's the modifications  I did to get things building on Windows with MinGW, with the outcome that with the correct environment setup it should be a case of just supplying these three commands:-

python waf configure --prefix="C:\Program Files (x86)\CodeBlocks\MinGW"

python waf

python waf install

-----------------
You need to install python and MinGW with pthreads (I used Codeblocks with built in TDM-GCC). During the configure stage it copies the dependencies into bin/include/lib in the MinGW root specified by the prefix option.

I took the built dependencies from the mingw_port and made a few changes:- 

- I removed the pthread headers from libav as TDM-GCC has them already.
- Recompiled libsamplerate to fix def file / dll inconsistency
- moved taglib headers down a level in /include/taglib and added missing "tnmap.tcc"


carthach:
I also got it working with MSVC somewhere, will have a look when I get a chance.

q-depot:
Hi,

I tried to compile your fork, but the build failed. I'm using the latest mingw with pthread.  Waf configure  seems fine.

Did you have a chance to look at the VS project? 

essentia

carthach:
Not yet, but by the looks of it it's failing due to the VS headers (versus GCC/MinGW). Maybe try mess around with your include path?

q-depot:
Can you please be more specific? I'm not too familiar with waf and mingw.

About VS I've managed to compile the static library in vs2012, I had to replace a bunch of functions, however the .lib file is somewhat huge, in debug is about 1GB and in release 500Mb, still can't understand what's included in the lib.

thanks.

carthach:
Aye,

I had the same big library file issue with VS, but when you compile it down into your end application it's typically a lot smaller, depending on what you use from the library (mine compiled down to 6MB).

Anyway, going back to your problem with MingGW, it seems to be using VS headers (judging from your paths). It should be using the MinGW headers (mine is C:\Program Files (x86)\CodeBlocks\MinGW\include), so put that before your path for the VS headers and it might work!

q-depot:
is this something in the wscript or a parameter for waf?

back to VS did you compile the static library or the dll?

carthach:
Actually by the looks of it, I did nothing. Just try and get a command prompt up without the Visual Studio paths set somehow.

With VS I compiled a static library.

q-depot:
I finally compiled the library .a however it doesn't work in VS2012, I get unresolve links, can it be used in VS?

carthach:
Not sure!

As the title says experimental support... I would imagine you'd be better off working in Codeblocks or something similar.

Also, bear in mind I've been using the ----lightweight=FFTW option to exclude all the extraneous stuff like ffmpeg which I didn't use..

C

q-depot:
I did use lightweight too, I guess this method works for CodeBlocks, I'll have another go with the VS project, do you mind if I'll have a look at your VS solution?

thanks.

carthach:
Course, you might find something helpful here:-
https://github.com/carthach/rhythmCAT/tree/master/Builds/VisualStudio2013

q-depot:
this is a project that use essentia, I thought you had a VS project to compile essentia as a static library on its own.

carthach:
Afraid not, I just used the python script.

AnaTP:
Hello everyone,

I'm new at essentia and I would like to use it in a visual studio if I could. @q-depot did you manage to build the library through VS? And did you manage to create a project to use essentia ?

I can always use the library with code CodeBlocks, are there any advances in compiling essentia in windows in an easier way?

Thanks in advance

Oeyvind:
I'm trying to compile with mingw64, but it seems it looks for msvc where I want it to look for gcc/g++ (?)

Administrator@HF-31335 MINGW64 /c/Source/essentia-master/essentia-master
$ python waf configure --prefix="C:\msys64\mingw64" --with-static-examples
Setting top to                           : C:\Source\essentia-master\essentia-master
Setting out to                           : C:\Source\essentia-master\essentia-master\build
→ configuring the project in C:\Source\essentia-master\essentia-master
→ Building in release mode
copying pkgconfig ...
copying yaml-0.1.5...
copying fftw-3.3.3...
copying libav-0.8.9...
copying libsamplerate-0.1.8...
copying taglib-1.9.1...
Checking for 'msvc' (c++ compiler)       : c:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\BIN\CL.exe
Checking for 'msvc' (c compiler)         : c:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\BIN\CL.exe
(...snip)

When building I get
...
algorithm.cpp
c:\source\essentia-master\essentia-master\src\essentia\types.h(23) : fatal error C1083: Cannot open include file: 'map': No such file or directory
...
and then a bunch of other errors

I also tried to add this to wscript, line 72, in function configure:
    ctx.find_program('gcc', var='CC', mandatory = True)
    ctx.find_program('g++', var='CXX', mandatory = True)

... which gives me 
Checking for 'msvc' (c++ compiler)       : Command ['C:\\msys64\\mingw64\\bin\\g++.exe', '/nologo', '/help'] returned 1

How come it insists on looking for the msvc compiler?

PR #250, opened 2015-04-13
Kiss and Accelerate FFT selector

carthach:
I noticed in the latest configure that there's talk of implementing a new FFT. I was tinkering around with FFTs last week and I've got essentia working with KISS and Accelerate framework on OS X. You can select which fft to use in the configure script like so:

./configure --fft=<OPTION>

Options being:-
FFTW (default) KISS (no external linking needed) ACCELERATE (native on Mac/iOS)

Basically I duplicated the FFTW.h/cpp to FFTK.h/cpp and FFTA.h/.cpp respectively (as well as the IFFTs). The waf script then takes care of choosing the appropriate cpp for building.

I haven't tested it much, I ran standard_mfcc example for each though and it seems to be working.

There's some work still to be done with the IFFTs need to fix some bugs and test them...


dbogdanov:
Nice!!!

carthach:
Should be good to go now!

Issue #179, opened 2014-10-21
different results for 32 bit builds (and minor differences between runs)


JonnyJD:
This is the result on a 64 bit build of v2.1_beta1-28-g21ef5f4 on Arch Linux:
https://gist.github.com/JonnyJD/fe9f07a8f48820307ee4#file-64bit_arch-json
_this is the *_base***

This is the 32 bit acousticbrainz-provied build (v2.1_beta1-28-g21ef5f4-dirty) on the same (64 bit Arch Linux) machine:
https://gist.github.com/JonnyJD/fe9f07a8f48820307ee4#file-32bit_provided-json
_this is the one *_off***
difference to _base_ (that is 64 bit vs. 32 bit):
https://gist.github.com/JonnyJD/fe9f07a8f48820307ee4#file-arch64_to_provided32-diff

This is the 64 bit acousticbrainz-provied build (v2.1_beta1-28-g21ef5f4-dirty) on the same machine:
https://gist.github.com/JonnyJD/fe9f07a8f48820307ee4#file-64bit_provided-json
difference to _base_:
https://gist.github.com/JonnyJD/fe9f07a8f48820307ee4#file-arch_to_provided-diff

The processed track is:
https://musicbrainzsoundteam.bandcamp.com/track/surprisingly-o-k-variations-on-this-shit
(free download, just set 0 $)

dbogdanov:
No need for `--no-msse` flag for non-x86 architectures since the https://github.com/MTG/essentia/commit/8e5652c8fad054843b844e08037866ae5828a9ee commit.

dbogdanov:
Closing this in favor of a more generic robustness evaluation (new issue).

PR #258, opened 2015-05-27
don't force SSE instructions on ARM

tcwalther:
The fix for https://github.com/MTG/essentia/issues/179 forces SSE instructions to make sure outputs on i386 and x86_64 are the same. However, this breaks compilation on ARM platforms, as SSE is an intel-specific instruction set.

This patch only forces SSE instructions when not on ARM. Theoretically, the correct way would probably be to check whether one is on an intel platform (rather than not on ARM), but there are multiple possibilities for intel hardware, and I don't have hardware to test them all. Furthermore, it seems like the result of platform.machine() can be i386, x86, x86_64, AMD64, and maybe something I've missed, so it's probably to check if it is not ARM.

### DIFF
--- a/wscript
+++ b/wscript
@@ -59,7 +59,9 @@ def configure(ctx):
     # force using SSE floating point (default for 64bit in gcc) instead of
     # 387 floating point (used for 32bit in gcc) to avoid numerical differences
     # between 32 and 64bit builds (see https://github.com/MTG/essentia/issues/179)
-    ctx.env.CXXFLAGS += [ '-msse', '-msse2', '-mfpmath=sse' ]
+    # Don't force SSE on ARM platforms, though
+    if not 'arm' in platform.machine().lower():
+        ctx.env.CXXFLAGS += [ '-msse', '-msse2', '-mfpmath=sse' ]
 
     # define this to be stricter, but sometimes some libraries can give problems...
     #ctx.env.CXXFLAGS += [ '-Werror' ]


alastair:
Before we merge this I'd like to test that we don't see the same differences in output on ARM that we saw between x86/x64 before we enabled SSE. What type of machine are you testing this on? I'll see if I can get something compiled on a rpi and do a quick test.

tcwalther:
I am testing this on an Raspberry Pi B+. I don't know if we will get differences in output, but from how I understand #179, the differences are basically due to different rounding on different floating point precisions. Forcing SSE instruction sets just enforces a certain rounding method on all chips that support it (rounding related to floating point precision, not mathematical rounding).

SSE is simply not available on ARM. If there are small output differences on ARM, I don't think there's anything one can do about it. At most, one could keep them equal across different ARM platforms, which would mean testing on ARMv6 vs ARMv7 (i.e., Raspberry Pi B/B+ vs Pi 2 B, and some Arduino platforms), but honestly that's a luxury problem when it doesn't even compile in the first place. We're talking about a cumulation of floating point precision differences here, which, even cumulated, are still very minor. Given that ARM and x86 are entirely different platforms, those differences should be expected.

tcwalther:
alastair, can you tell me which tests to run to help merge the pull requests?

alastair:
Hi Thomas,
Sorry - I haven't completely forgotten this.
We have some scripts at https://github.com/MTG/acousticbrainz-client/tree/master/tools which we were using to look at the results of the same file generated on different architectures (or the same recording with different codecs, etc).
We don't have anything much more formal yet - the plan is to compute acceptable tolerances by running this type of test on many different files, and then develop a system to automatically ensure that changes/new platforms do not fall outside these levels, however this work hasn't been done yet.

Alastair

PR #264, opened 2015-06-10
New algorithms from MusicBricks project for MHD workshop

jjaner:


dbogdanov:
There is a huge memory leak somewhere in the new algorithms. Have you checked running them from python? This code hangs my machine forever:
```
import essentia.standard as esd
```
You can test this building a new branch https://github.com/MTG/essentia/tree/musicbricks_merge

jjaner:
I have not checked the algorithms on Python. It was tested only as a standalone binary exampl on Linux/Ubuntu14.

dbogdanov:
This branch was originally forked from 2.0.1 and there was a mess merging it to master. I've redone this merge by cherry-picking the commits. As for now, I cannot reproduce the system freeze on my machine.

dbogdanov:
If we plan to work on further updates, it would be the best to re-fork the master branch to avoid further possible merge conflicts.

jjaner:
So, if I get it correctly, branches 2.0.1 and 2.1 from essentia repository diverged, and the 'margin mess' was due to that  essential-musicbricks was originally forked from 2.0.1. Is that correct?
you are right. Since we plan to do further update  until end of the project, we should re-fork from master branch.

PR #278, opened 2015-08-07
Can now build with Emscripten targeting Javascript

rhaps0dy:
I should probably add instructions on how to build with Emscripten. How can I do that?


dbogdanov:
Yes, we need instructions. They can go to [FAQ.md](https://github.com/MTG/essentia/blob/master/FAQ.md)

PR #279, opened 2015-08-07
Separated Network::run into runPrepare and runStep

rhaps0dy:
This way one can only sort the network once, and then run just a step in
the analysis and analyze audio in real time.

It might be useful to include the algorithms I wrote that just generate input from a buffer in memory (so the buffer can change with new audio data and runStep can be called again) and an algorithm that outputs to a given place in memory, for the same reason.

### DIFF
--- a/src/essentia/scheduler/network.cpp
+++ b/src/essentia/scheduler/network.cpp
@@ -220,6 +220,15 @@ bool algorithmHasProduced(Algorithm* algo, EssentiaMap<SourceBase*, int>& produc
 }
 
 void Network::run() {
+  runPrepare();
+  while (runStep());
+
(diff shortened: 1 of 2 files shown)


dbogdanov:
How the new algorithms differ from [RingBufferInput](http://essentia.upf.edu/documentation/reference/streaming_RingBufferInput.html) and [RingBufferOutput](http://essentia.upf.edu/documentation/reference/streaming_RingBufferOutput.html)?

rhaps0dy:
Only in implementation! Their purpose appears to be the same. I don't quite understand how to use the RingBuffer algorithms, I guess it boils down to calling "add" whenever you have new data?

I'll try and add some documentation for it.

PR #287, opened 2015-08-27
Fixed src/wscript to install the correct (generated) pkgconfig file...

ChristianFrisson:
…and removed the unnecessary (static) one from the repo root.

Spotted it when trying to install Essentia in MacPorts whose default prefix is /opt/local, the pkgconfig file was always stating /usr/local...

### DIFF
--- a/wscript
+++ b/wscript
@@ -183,7 +183,7 @@ def configure(ctx):
     Name: libessentia
     Description: audio analysis library -- development files
     Version: %(version)s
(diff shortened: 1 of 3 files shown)


wscript
```diff
@@ -183,8 +183,8 @@ def configure(ctx):
     Name: libessentia
```
alastair: The mode of this file shouldn't have been changed to 755

dbogdanov: Fixed

ChristianFrisson: Thank you, sorry for my mistake!

PR #296, opened 2015-09-09
Pitchfiltermakam

sertansenturk:
Here are the changes related to parametrizations. 

Also I realised that we were not initialising the confidence values correctly and fixed it. (2eb63db)


alastair:
merged, let's continue discussion on the original PR

PR #298, opened 2015-09-11
move ringbuffer input algorithms 

MartinHN:


MartinHN:
allows to get ringbuffer special function (add / get) without including weird search path when using ringbuffer input/output algorithms

PR #305, opened 2015-09-22
Android build

carthach:
This adds support for compiling on android.

An example of how to build using a lightweight version of Android (with the KISS FFT) is in the 
./build_android.sh

Update the PATH to point to where you have your Android Standalone Toolchain:-
https://developer.android.com/ndk/guides/standalone_toolchain.html

I would suggest creating a customised toolchain (as suggested in on that webpage) using the make-standalone-toolchain.sh script.

### DIFF
--- a/build_android.sh
+++ b/build_android.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+export PATH=~/Dev/android/toolchain/bin:$PATH
+./waf clean
+./waf configure --cross-compile-android --lightweight= --fft=KISS --ignore-algos=LPC;
+./waf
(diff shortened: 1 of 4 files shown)


carthach:
Hmmm, maybe just get rid of that shell script altogether then, that comment should be sufficient.

Some declaration in the LPC threw an error, maybe some conflict with the Android headers, I'll look further. The prefix can be ignored as well, that was just for me...

carthach:
The uppercase _P in LPC conflicts with a macro in the headers for the Android toolchain I think, I changed it to lowercase and it works now.

PR #331, opened 2015-12-11
added new key profiles and some basic description

angelfaraldo:


angelfaraldo:
I don't think I touched any notation, did I? Just added very short descriptions of the context of each profile and added one refference

angelfaraldo:
so how to merge this with onto the official essentia?

dbogdanov:
@angelfaraldo I mean, we should use "A#" instead of "Bb", etc.

angelfaraldo:
renamed!

dbogdanov:
Thank you!!

PR #352, opened 2016-01-19
Fix memory leak in RhythmExtractor2013

jlyda:
I found a big memory leak for the RhythmExtractor2013 which was related with the missing deletion of referenced algorithms in OnsetDetectionGlobal. Also RhythmExtractor2013 and BeatTrackerMultiFeature do not apply correctly the network->clear() principles.

### DIFF
--- a/src/algorithms/rhythm/onsetdetectionglobal.h
+++ b/src/algorithms/rhythm/onsetdetectionglobal.h
@@ -82,6 +82,11 @@ class OnsetDetectionGlobal : public Algorithm {
     if (_frameCutter) delete _frameCutter;
     if (_windowing) delete _windowing;
     if (_spectrum) delete _spectrum;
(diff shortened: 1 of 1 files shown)


dbogdanov:
Thanks! No need to call ```_network->clear();``` instead of ```delete _network;``` because the ```clear()```method is already called in destructor. Please, change back before the merge.

jlyda:
Thanks a lot! That's true. But profiling the memory consume I had a lower memory footprint with clear than with the delete approach. Maybe it is related with my application. I will give it a deeper look.

jlyda:
I changed it back to delete. It has the same memory footprint. Thanks a lot. I will commit and merge then.

jlyda:
Ok, I fixed it. Unfortunately I do not have write access permission. Can you merge it Dmitry?

PR #374, opened 2016-02-10
GFCC inputsize parameter

jlyda:
I added inputSize parameter with default value 1025 as it is set in ERBBands.
Updated a cout to E_INFO std output.


dbogdanov:
Thanks!

PR #447, opened 2016-07-27
Remove FFTWComplex from Accelerate and KISS FFTs for now

carthach:
This new FFTWComplex algorithm needs to be adapted as FFTAComplex (for Accelerate) and FFTKComplex for (KISS) otherwise it'll throw errors.

This pull request will remove the algorithm when you use these FFTs for now.


dbogdanov:
I've renamed FFTWC to a generic name "FFTC"

PR #460, opened 2016-08-23
Modify the oddtoevenharminicenergyratio algorithm

hellska:
Set the maximum value for the ratio to 1000(arbitrary value)
Set the ratio to 1 when the even energy is zero and the odd is less than 0.01(arbitrary value)

### DIFF
--- a/src/algorithms/tonal/oddtoevenharmonicenergyratio.cpp
+++ b/src/algorithms/tonal/oddtoevenharmonicenergyratio.cpp
@@ -68,14 +68,18 @@ void OddToEvenHarmonicEnergyRatio::compute() {
     else           odd_energy += magnitudes[i] * magnitudes[i];
   }
 
-  if (even_energy == 0.0) {
-     oddtoevenharmonicenergyratio = numeric_limits<Real>::max();
+  if (even_energy == 0.0 && odd_energy > 0.01) {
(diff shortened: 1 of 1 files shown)


hellska:
I set the threshold of odd harmonics to 0.01 because this energy value is inaudible for our hearing system. Still not sure of the 1000 value, maybe a ratio that represent a difference of 80dB instead of an arbitrary value can be more reasonable. Any suggestion is welcome.

dbogdanov:
Yes, 80dB would be fine

src/algorithms/tonal/oddtoevenharmonicenergyratio.cpp
```diff
@@ -68,14 +68,18 @@ void OddToEvenHarmonicEnergyRatio::compute() {
...
+  }
+  else if (even_energy == 0.0 && odd_energy < 0.01 ) {
+     oddtoevenharmonicenergyratio = 1;
   }
   else {
      oddtoevenharmonicenergyratio = odd_energy / even_energy;
   }
   if (oddtoevenharmonicenergyratio >= 1000.) {
-    cerr << "DEBUG: clipping oddtoevenharmonicenergyratio to max value" << endl;
+    cerr << "DEBUG: clipping oddtoevenharmonicenergyratio to max value threshold" << endl;
```
dbogdanov: Use E_WARNING("...") instead

PR #463, opened 2016-09-01
Loop BPM estimation algorithms

ffont:
This pull request implements the Percival's BPM estimation method as described in [1] and a confidence measure for BPM estimates of music loops as described in [2].
A number of new algorithms have been added:

- **`PercivalBpmEstimator`**: takes as input an audio signal and outputs a BPM estimate.
- **`LoopBpmConfidence`**: takes as input an audio signal assumed to be a music loop and a BPM estimate and outputs a confidence value [0,1].
- **`LoopBpmEstimator`**: combines the previous two algorithms. Takes as input an audio signal assumed to be a music loop and is configured with a `confidenceThreshold` parameter. It estimates the BPM using `PercivalBpmEstimator` and computes a confidence value with `LoopBpmConfidence`. If computed confidence is above the threshold, the estimated BPM is outputted, otherwise the algorithm outputs 0.

To implement the algorithms above, another number of algorithms have had to be added/modified:

- **`EnhanceHarmonics`**: implements a required step of `PercivalBpmEstimator`.
- **`EvaluatePulseTrains`**: implements a required step of `PercivalBpmEstimator`.
- **`Autocorrelation`**: we added an option to compute generalized autocorrelation as described in [3] (configuring autocorrelation with `generalized` parameter, see `PercivalBpmEstimator`).
- **`UnaryOperator`**: we added `scale` and `shift` parameters to mimic `UnaryOperatorStream`.
- **`Windowing`**: we added a `normalized` parameter which defaults to `true` (so we don't change default window behaviour) but it can be set to `false` and then no normalisation is applied to the computed window.

[1] Percival, G., & Tzanetakis, G. (2014). Streamlined tempo estimation based on autocorrelation and cross-correlation with pulses. IEEE/ACM Transactions on Audio, Speech, and Language Processing, 22(12), 1765–1776.

[2] Font, F., & Serra, X. (2016). Tempo Estimation for Music Loops and a Simple Confidence Measure. Proceedings of the International Society for Music Information Retrieval Conference (ISMIR).

[3] Tolonen T., and Karjalainen, M. (2000). A computationally efficient multipitch analysis model.  IEEE Transactions on Audio, Speech, and Language Processing, 8(6), 708-716.


dbogdanov:
```EnhanceHarmonics``` sound too generic and one can be easily mislead on what it does. May be change to ```PercivalEnhanceHarmonics```.

dbogdanov:
The same for ```EvaluatePulseTrains```

dbogdanov:
Furthermore, when presented a sorted list of algorithms in the documentation overview it will be clear that they are all related.

ffont:
Sound good, I'll do it!

PR #475, opened 2016-09-29
Add meta tag for improved responsiveness

giuband:


giuband:
The same tag should be added in the algorithms pages

PR #495, opened 2016-11-02
SpectrumToCent Algorithm

pabloEntropia:
This is an implementation of the SpectrumToCent proposed algorithm. The algorithm creates an array of frequencies with a fix resolution on the cent scale and then uses it to call [TriangularBands](http://essentia.upf.edu/documentation/reference/std_TriangularBands.html). It returns an array with the energy of each new bin and an array with their central frequencies. 
Input:
- spectrum

Outputs:

- bands: the energy in each band

- frequencies: the central frequency of each band

 
Parameters are:
- sampleRate

- log: compute log-energies (log10 (1 + energy))

- minimumFrequency: central frequency of the first band of the bank [Hz]

- centBinResolution: Width of each band in cents. Default is 10 cents

- bands: number of bins to compute. Default is 720 (6 octaves with the default 'centBinResolution'


dbogdanov:
Great. Please add streaming mode ([example](https://github.com/MTG/essentia/blob/master/src/algorithms/spectral/triangularbands.h#L70)).

dbogdanov:
Closing this pull request as we will have another one including other updates too.

Issue #494, opened 2016-10-31
Faster TriangularBands implementation


dbogdanov:
Currently, the triangular weighting for spectrum frequency bins is computed on the fly inside compute(). We should consider precomputing filter weights just as it is done in [MelBands algorithm](https://github.com/MTG/essentia/blob/master/src/algorithms/spectral/melbands.cpp#L74). This would lead to faster computation but larger memory consumption in the case of very large number of bands and spectrum bins. 

MelBands can be later re-implemented to re-use TriangularBands instead of duplicating the code later on.

georgid:
@pabloEntropia  
In the TriangularBands we should add a boolean parameter `normalize` that signifies if the default filter height of 1 is normalized or not.  
Also in MelBands the formula for the warping function from spectrum to mell should be given as a parameter.

dbogdanov:
We can have two normalization types for ```normalize```: "unit_sum" and "unit_max".

pabloEntropia:
I have been working on the reimplementation on TriangularBands as basis for all the methods relaying on triangular filterbanks. I've found another feature that we should consided. From [MelBands](https://github.com/MTG/essentia/blob/master/src/algorithms/spectral/melbands.cpp) the triangles are computed on the mel scale, not only the boundary frequencies of it (as would happen on the current implementation of TriangularBands).    
 `for (int j=jbegin; j<jend; ++j) {
      Real binfreq = j*frequencyScale;
      // in the ascending part of the triangle...
      if ((binfreq >= _filterFrequencies[i]) && (binfreq < _filterFrequencies[i+1])) {
        _filterCoefficients[i][j] = (hz2mel(binfreq) - hz2mel(_filterFrequencies[i])) / fstep1;
      }
      // in the descending part of the triangle...
      else if ((binfreq >= _filterFrequencies[i+1]) && (binfreq < _filterFrequencies[i+2])) {
        _filterCoefficients[i][j] = (hz2mel(_filterFrequencies[i+2]) - hz2mel(binfreq)) / fstep2;
      }
}`
In order to fix this I would need to create a new parameter defining the scale for the weight computation. So far, and also considering #501, the parameter list  in order to make the function as flexible as possible would be:
[current parameters]
- sampleRate
- frequencyBands
- log
[new parameters]
- "inputSize". Default 1025 ( in order to attempt to compute the weight on the Configure() , otherwise they will be computed on the first iteration of Compute())
[...]

PR #505, opened 2016-11-09
Faster triangular bands

pabloEntropia:
This request contains all my changes so far. New implementation of TriangularBands acts as proposed in #494. The function passes the current tests, however I would like to update them in order to test also the new parameters. 
I set a default `inputSize` of 65536 to prevent it to break in the configure step when asking for very narrow bands.


dbogdanov:
Closing this pull request as we will have another one including other updates too.

Issue #518, opened 2016-11-24
Finalize work on TriangularBands, MelBands, and MFCCs


dbogdanov:
@pabloEntropia  
- Now I see there are various new branches in your repository. Please merge them all together into one and make a new pull request (or merge them all into one existing pull request) otherwise it is difficult to track things.
- Update unit tests for TriangularBands testing new parameters 
- Add a few "real-case" regression tests in MelBands unit test (use a few of different spectrum inputs and default parameters. Use the output from older version of MelBands as a reference ground truth). This is to make sure that results of MelBands are identical after code updates. 
- Add another "real-case" regression test (or few of them) comparing htk MelBands using results from htk as a reference.

@georgid

pabloEntropia:
I have created a branch merging all my work here in order to make it possible to track all my commits. 
I will create the pull request when I am satisfied with the results of all the tests.
[TriangularBands-tools](https://github.com/pabloEntropia/essentia/tree/TriangularBands-tools)

dbogdanov:
You've also got some work in faster-TriangularBands. Is that to be merged too?

pabloEntropia:
Yes it is!

PR #520, opened 2016-11-28
Triangular bands tools

pabloEntropia:
This request summarizes all my changes so far plus the proposals on  on #518:
- TruangularBands: I added the new parameters to the unit test. I also tested that the power spectrum matches the magnitude bands for the unit spectrum case.
- MelBands: I left the regression test as it was to test that if you don't specify any parameter, the default configuration matches the former behavior. I included another regression test, using an HTK output provided by @georgid to test the "htk styled" Mel bands 

Things to do:
- I've seen that MFCC test has one error (because of the last TriangularBands exception added) and one failure, however as the MelBands  is passing the old  regression test is  with the old parameters my changes shouldn't have affected it however I will try to understand why and repair it.
- I've to fix the normalization of DCT type III in order to behave like Librosa


dbogdanov:
The MFCC regression test was failing for some time already. Create a new regression test for MFCC from scratch.

pabloEntropia:
I have fixed the errors on  the MFCC unit test and created a new regression test  for the new HTK style implementation comparing with the MFCC calculated by Librosa. Should I remove the old regression test?

test/src/unittest/spectral/test_mfcc.py
```diff
@@ -38,23 +39,58 @@ def testRegression(self):
...
+            bands, mfcc = self.InitMFCC(size)(ones(1025))
             self.assertEqual(len(mfcc), size )
             self.assertEqual(len(bands), 40 )
             self.assert_(not any(numpy.isnan(mfcc)))
             self.assert_(not any(numpy.isinf(mfcc)))
             size -= 1
 
+    def testRegressionHtkMode(self):
+        audio = essentia.array(np.loadtxt(join(filedir(), 'mfcc', 'audio.txt')))
+        trueMFCC = np.loadtxt(join(filedir(), 'mfcc', 'true_mfcc.txt'))
```
dbogdanov: Better file name: "true_mfcc_htk.txt"?

dbogdanov:
The ```test/src/unittest/spectral/mfcc/audio.txt``` and ```true_mfcc``` files are missing.

dbogdanov:
Please add these files before I do the merge.

dbogdanov:
We'll have to fix regression test for htk MFCC. New test data is too large (33M for encoding 30 seconds of audio stored in plain text  + 1.5M for reference MFCC values in frames). This is not a good practice to store that in our github repository if we can avoid that.

Do this instead:
- [remove the last commit from git history in local and remote](http://stackoverflow.com/a/8225166/603642) (ef5e616) (be careful!)
- do a new commit with updated test
    - use a test audio file from ```test/audio/recorded/``` with MonoLoader
    - pre-compute HTK reference values, however, do not store frames, but only the mean. These would be only 20 values, you can store them in the code itself.

PR #538, opened 2016-12-15
Liftering and DCT III scaling to fit HTK

pabloEntropia:
We have already tested that the new parameters behave the same as HTK. Now I should update the MFCC and DCT tests  against the HTK MFCCs and document how to use Essentia in the HTK way.


pabloEntropia:
Tests are working again now. [Here](https://github.com/georgid/mfcc-htk-an-librosa/blob/master/MFCC-htk-with-liftering-real-case-Test.ipynb) you can see the difference between Essentia and HTK with all the considerations so far. I have documented how to replicate HTK inside the MFCC code but I couldn't properly compile the documentation to see if everything is correct. Maybe you would like to check it before merging @dbogdanov

dbogdanov:
Fixed doc formatting and merged. Thanks!

Issue #487, opened 2016-10-20
implement inverse 12 MFCCs for visualisation 


georgid:
The first MFCCs are standard for describing singing voice timbre. The MFCC feature vector however does not represent visually well the singing voice. Instead, a common practice is to use for visualisation the inverted first some (12-15) MFCC coefficients to mel-bands domain.
Port from the invmelfcc.m  as explained here http://labrosa.ee.columbia.edu/matlab/rastamat/
or from python: 
https://github.com/avaitla/HenryVsRudolph/blob/master/Python/MFCC.py

dbogdanov:
The algorithm should have a parameter to specify how many coefficients to take from the MFCC vector.

georgid:
implementation plan: 
1) output the _dctTable from the DCT in essentia
https://github.com/MTG/essentia/blob/528c57494ae3322ffed4442f44082ef0608c0ee8/src/algorithms/standard/dct.cpp
2) invert it, use MEL_BANDS  and a parame: MFCC_COEFS: how many coefficients to take from the MFCC
invD = numpy.inv(_dctTable)
3) multiply MFCC * invD

georgid:
@pabloEntropia 
Once you are done, may you add in both in the documentation of InverseDCT and MFCC that inverseDCT can be used to compute smoothedMelBands in the following way: 
1) compute MFCC 
2) smoothedMelBands = inverseDCT(computed MFCC)

PR #540, opened 2016-12-21
Inverse DCT implementation according to #487

pabloEntropia:
- Added some regression tests based on DCT tests
- supports DCT II and DCT III inversion
- supports de-liftering
- documentation on how to perform inverse MFCC added


src/algorithms/standard/idct.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+
+void IDCT::compute() {
+
+  const vector<Real>& input = _dct.get();
+  vector<Real>& idct = _idct.get();
+
+  vector<Real> dct = input;
+
```
pabloEntropia: Here I copied the input to a new vector in order to be able to use it. Is there a faster way to do it? @dbogdanov

dbogdanov: That's good

src/algorithms/standard/idct.cpp
```diff
@@ -29,9 +29,10 @@ const char* IDCT::category = "Standard";
 const char* IDCT::description = DOC("This algorithm computes the Inverse Discrete Cosine Transform of an array.\n"
 "It can be configured to perform the inverse DCT-II form, with the 1/sqrt(2) scaling factor for the first coefficient or the inverse DCT-III form based on the HTK implementation.\n"
 "\n"
-"IDCT can be use to compute smoothed Mel Bands. In order to do this: \n "
+"IDCT can be used to compute smoothed Mel Bands. In order to do this: \n "
 "- compute MFCC\n"
 "- smoothedMelBands = 10^(IDCT(MFCC)/20) \n"
+"Note: The second step assumes that the default 'logType' were used to compute MFCCs, otherwise that formula should be changed in order to be consistent.\n"
```
dbogdanov: explicitly write the 'logType' value  just in case we will change default value in future

src/algorithms/spectral/mfcc.cpp
```diff
@@ -60,6 +60,11 @@ const char* MFCC::description = DOC("This algorithm computes the mel-frequency c
 "\n"
 "This algorithm depends on the algorithms MelBands and DCT and therefore inherits their parameter restrictions. An exception is thrown if any of these restrictions are not met. The input \"spectrum\" is passed to the MelBands algorithm and thus imposes MelBands' input requirements. Exceptions are inherited by MelBands as well as by DCT.\n"
 "\n"
+"IDCT can be used to compute smoothed Mel Bands. In order to do this: \n "
+"- compute MFCC\n"
+"- smoothedMelBands = 10^(IDCT(MFCC)/20) \n"
+"Note: The second step assumes that the default 'logType' were used to compute MFCCs, otherwise that formula should be changed in order to be consistent.\n"
```
dbogdanov: explicitly write the 'logType' value just in case we will change default value in future

PR #555, opened 2017-01-17
Check if an array is a view or a copy

pabloEntropia:
Modification in the python parser in order to check if an input numpy array is a view or a copy. If it is a view, it copies the data to a new array. 
There is also a base test in order to check the proper behavior.

### DIFF
--- a/src/python/essentia/standard.py
+++ b/src/python/essentia/standard.py
@@ -20,7 +20,7 @@ import essentia
 import common as _c
 import sys as _sys
 from _essentia import keys as algorithmNames, info as algorithmInfo
-
+from copy import copy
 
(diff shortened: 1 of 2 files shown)


dbogdanov:
https://docs.scipy.org/doc/numpy/reference/generated/numpy.ascontiguousarray.html

dbogdanov:
Use C_CONTIGUOUS flag instead of OWNDATA

seffka:
Is this fix aimed for all Algorithms?
Have an issue with standard.Slicer algorithm and sliced arrays. After upgrade to latest dev version it's still here.

dbogdanov:
Should be fixed now.

Issue #552, opened 2017-01-10
AudioLoader: add support for audio sources with multiple audio streams


louis49:
I got these error when loading an audio file containing multiple audio streams :
AudioLoader ERROR: found 4 streams in the file, expecting only one audio stream

Do you know how to get only first/nth/last stream ?

dbogdanov:
Currently there is no way to specify the audio stream. You can manually extract the streams using ffmpeg. However this would be a very useful option to have.

dbogdanov:
Add new parameter ```audioStream``` (audio stream index) to AudioLoader. Default value = 1 corresponds to the first audio stream among the streams in the source.

https://github.com/MTG/essentia/blob/master/src/algorithms/io/audioloader.cpp#L104

dbogdanov:
Throw exception if there is no audio streams.

dbogdanov:
@pabloEntropia

dbogdanov:
Fixed in #557

PR #557, opened 2017-01-20
Update of load algorithms to support multistreams #552

pabloEntropia:
The update includes `AudioLoader`, `Monoloader` and  `EasyLoader`. The streaming test has also been updated.

### DIFF
--- a/src/algorithms/io/audioloader.h
+++ b/src/algorithms/io/audioloader.h
@@ -63,6 +63,8 @@ class AudioLoader : public Algorithm {
   struct AVAudioResampleContext* _convertCtxAv;
 
   int _streamIdx; // index of the audio stream among all the streams contained in the file
+  std::vector<int> _streams;
+  int _selectedStream;
   bool _configured;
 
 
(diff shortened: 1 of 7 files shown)


src/algorithms/io/audioloader.h
```diff
@@ -172,6 +175,7 @@ class AudioLoader : public Algorithm {
   void declareParameters() {
     declareParameter("filename", "the name of the file from which to read", "", Parameter::STRING);
     declareParameter("computeMD5", "compute the MD5 checksum", "{true,false}", false);
+    declareParameter("audioStream", "audio stream index to be loaded. Other streams are no taken into account (e.g. if stream 0 is video and 1 is audio use index 0 to access it.)", "[0,inf)", 0);
```
alastair: Other streams are no*t* taken

test/src/unittest/io/test_audioloader_streaming.py
```diff
@@ -240,6 +247,40 @@ def testMD5(self):
         self.assertEqual(md5_mp3, "1e5a598218e9b19cfe04d6c2f61f84a6")
         self.assertEqual(md5_ogg, "a87dad40fea0966cc5b967d5412e8868")
         self.assertEqual(md5_aac, "9a4c7f0da68d4b58767f219c48014f9c")
+    
+    def testMultiStream(self):
+        
+        # A multistream file containing the same audio on streams 0 and 1 
+        filename = 'multistream.mka'
```
alastair: I would reconsider this test - it would also succeed if the audioStream parameter was ignored and both AudioLoaders opened the first stream. Is it possible to have two multistream files, where one has streams a + b and the other b + a?

PR #560, opened 2017-02-08
NSGConstantQ

pabloEntropia:
This is my draft of the implementation of the Non Stationary Gabor ConstantQ transform. It is not working yet. I put it here just in case you want to have a look on the code @dbogdanov


pabloEntropia:
I have a first implementation consistent with the MATLAB implementation. However there are still a lot of things to improve. You can see the current look of the transform in this notebook @dbogdanov   
https://github.com/pabloEntropia/CQ-transform-master/blob/master/CQ%20development.ipynb

pabloEntropia:
Essentia IFFTs returns a real vector, which is OK when you deal  Hermitian Functions (which is the case most of the time). However, it is not the case for the frequency filters I'm dealing with, so the output of my IFFT should be complex. I've been trying to implement a FFTW3 based complex IFFT but I find it too time-consuming, so I wonder if you could give me some indications for this (or maybe it would be easier using another library) @dbogdanov

dbogdanov:
:clap:

dbogdanov:
So, it currently works only with FFTW, but that's ok for now. 

We also support KissFFT library and Accelerate on Mac for FFT,  and we'll need to add support for that too. KissFFT has an appropriate[method](https://github.com/itdaniher/kissfft/blob/master/tools/kfc.h#L37) for complex inverse transform (KissFFT source code is included in Essentia). 

Accelerate [too](https://developer.apple.com/reference/accelerate/1450224-vdsp_fft_zip).

pabloEntropia:
[Update] 
After fixing some errors I have achieved the same reconstruction noise levels than the MATLAB implementation.  I've also created a streaming wrapper but the Essentia Pool is not accepting complex matrices as input at the moment so I will try to add support for them now. Some tests and reconstruction examples can be seen in this[notebook](https://github.com/pabloEntropia/CQ-transform-master/blob/master/CQ%20development.ipynb)

pabloEntropia:
I added the proposed modifications and created this [notebook](https://gist.github.com/pabloEntropia/ce228ade1c6acf8e2d8926b447e8eb98) explaining the algorithm with some examples.

dbogdanov:
Can you add the notebook to the ```src/examples/python/```? So that I will be able to comment on it in the pull request review.

pabloEntropia:
Done!

dbogdanov:
A comments on the notebook: can one use [OverlapAdd](http://essentia.upf.edu/documentation/reference/std_OverlapAdd.html) algorithm for reconstruction?

dbogdanov:
I've improved the notebook in a new branch:
https://github.com/MTG/essentia/blob/pabloEntropia-NSGConstantQ/src/examples/python/NSGConstantQ-example.ipynb

dbogdanov:
Some other questions that we can comment on in the notebook:
- how does frame size affect the results in streaming mode? (the plots for standard and streaming are different)
- which window type to choose?

pabloEntropia:
This [paper](http://www.univie.ac.at/nonstatgab/pdf_files/dogrhove12_amsart.pdf) explains how to obtain the same representation using analysis but I find it too tricky to be implemented in an example notebook.

I tried to use OverlapAdd in the notebook but it doesn't do the same thing. In the example, frames are added as vectors. OverlaAdd adds all the elements in frame and returns an scalar each hopSize.

PR #572, opened 2017-03-09
Fix pitchyin

dbogdanov:


src/algorithms/tonal/pitchyin.cpp
```diff
@@ -89,7 +89,7 @@ void PitchYin::compute() {
   // Compute difference function
   for (int tau=1; tau < (int) _yin.size(); ++tau) {
     _yin[tau] = 0.;
-    for (int j=0; j < (int) _yin.size(); ++j) {
+    for (int j=0; j < (int) _yin.size()-1; ++j) {
```
alastair: unusual way of doing a for loop. Why not `j <= _yin.size()`?

dbogdanov: This is intentional to not over-go the signal size. The maximum index that can be achieved in j + tau should be less then framesize.

PR #597, opened 2017-05-02
Change Python library lookup from buildtime to runtime

jpauwels:
This should avoid segfaults when building the Python extensions with one Python version  (e.g. Homebrew Python) and using it with another one (e.g. System Python)


jpauwels:
Wait, I messed something up. The patch has the right idea, but probably my lack of knowledge about waf is the problem. The '=' definitely has to be a '+=', but that still doesn't work. I'll try to improve it later.

jpauwels:
Got it, shouldn't have used a space in the linker flag :-) This fixes my problems with the Python bindings and apparently I'm not the only one who's ran into it.

PR #623, opened 2017-07-09
Update docs look

giuband:
Make it look consistently with the homepage by sharing the same base templates.
Updated the Applications page.

Base templates for the documentation have been rewritten following modern accessibility standards.


doc/sphinxdoc/FAQ.rst
```diff
@@ -0,0 +1,584 @@
+Frequently Asked Questions
```
giuband: I actually have no clues on where this file comes from

PR #624, opened 2017-07-10
Update wscript to work on Python 2.6.6/CentOS 6

Jamie0:
Bug-fix to build on CentOS 6, where native Python support only goes up to 2.6.6 and doesn't support the set syntax.


Jamie0:
(I know that Python 2.7.0+ is supported, but this allows building without Python on CentOS 6)

PR #628, opened 2017-07-20
Only link against libessentia in the pc file. Fixes #340

### LINKED ISSUE #340: pcfile contains -lgaia2 even if gaia isn't configured
https://github.com/MTG/essentia/blob/67d76d66f5faff2182e99f47142109b0d5cb7d71/wscript#L205

alastair:
When we want to link libessentia.so into a program we only need to
specify -lessentia. Other dependencies of essentia are already
correctly linked against the .so. If a user wants to also link
their program against the same dependency (e.g. libav) then they
should explicitly link against that themselves.

### DIFF
--- a/wscript
+++ b/wscript
@@ -243,7 +243,7 @@ def configure(ctx):
     Name: libessentia
     Description: audio analysis library -- development files
     Version: %(version)s
-    Libs: -L${libdir} -lessentia -lgaia2 -lfftw3 -lyaml -lavcodec -lavformat -lavutil -lavresample -lsamplerate -ltag -lfftw3f
+    Libs: -L${libdir} -lessentia
     Cflags: -I${includedir}/essentia -I${includedir}/essentia/scheduler -I${includedir}/essentia/streaming -I${includedir}/essentia/utils
     ''' % opts
 


dbogdanov:
Those dependencies are required for the static version of the library. Probably two different .pc files can be generated depending on whether the shared or static libessentia was built.

sildeag:
One thing to think about is the location of the.pc files.  On `Ubuntu`, to the best of my understanding, `/usr/local/share/pkgconfig` is automatically searched.  `/usr/local/lib/pkgconfig` may be on the automatic search path in many other flavours of `linux` but requires `PKG_CONFIG_PATH` to be set to it on `Ubuntu`.

dbogdanov:
The correct way would be to specify Essentia's dependencies in [Libs.private](https://people.freedesktop.org/~dbn/pkg-config-guide.html) when generating the pc file.

@sildeag Here is some [related info](https://unix.stackexchange.com/questions/422869/why-are-pc-files-not-listed-in-one-place): ```/usr/share``` is for architecture-independent data, for that reason using ```/usr/local/lib/pkgconfig``` is more appropriate.

sildeag:
@dbogdanov thanks for the info.   I was just pointing out that `/usr/local/lib/pkgconfig` wasn't in the automatic search path for `Ubuntu.`  This impacts build scripts which use `.pc` files - since on `Ubuntu` not everyone has PKG_CONFIG_PATH set and/or has some out-dated `.pc` file on their system that is in the automatic search path.

PR #666, opened 2017-12-03
Ported all python files from python 2 to 3

colonder:


khiner:
PR 666! 😈

dbogdanov:
Great! :clap:  :smiling_imp: 
We've recently done the same porting too. I've just pushed the changes. Perhaps, we can cherry-pick some of your changes... The ones related to doc generation, for example.

colonder:
Oh sorry, I downloaded the beta3 version, maybe that's the cause that I didn't see that you've actually made the same thing. You should write some info in the releases tab that highlights that there's a newer version than beta3 because beta4 is completely invisible compared to beta3 big bold headline.

PR #687, opened 2018-03-07
StartStopCut

pabloEntropia:
An algorithm to detect cuts in the beginning/end of the audio plus unit tests. 

The algorithm works by iterating through audio frames until it reaches the first non-silent frame. If the first non-silent frame occurs before a configurable expected silent time (default 10 ms) the algorithm activates the start cut flag. Then the same is done backwards starting from the end. 
The benefit of this approach is that the algorithm doesn't have to iterate the whole signal as in  StartStopSilence. The drawback is that the streaming mode requires the use of RealAccumulator.


src/algorithms/audioproblems/startstopcut.cpp
```diff
@@ -0,0 +1,118 @@
...
+using namespace std;
+
+namespace essentia {
+namespace standard {
+
+const char* StartStopCut::name = "StartStopCut";
+const char* StartStopCut::category = "Audio Problems";
+const char* StartStopCut::description = DOC(
+    "This algorithm outputs if there is a cut at the beginning or at the end "
+    "of the audio by measuring the first and last non-silent frames and "
```
dbogdanov: "...by locating..."?

src/algorithms/audioproblems/startstopcut.cpp
```diff
@@ -0,0 +1,118 @@
...
+    "comparing them to the actual beginning and end of the audio. If the first "
```
dbogdanov: "comparing their positions..."

dbogdanov: What about something like this? 

"The input audio is considered to be cut at the beginning (or the end) and the corresponding flag is activated if the first (last) non-silent frame occurs before (after) the configurable time threshold."

src/algorithms/audioproblems/startstopcut.cpp
```diff
@@ -0,0 +1,118 @@
...
+    "non-silent frame occurs before a time threshold the beginning cut flag is "
+    "activated. The same applies to the stop cut flag.\n"
+    "\n"
+    "Notes: This algorithm is designed to operate file-wise. Use it in "
```
dbogdanov: "... to operate on the entire (file) audio"

src/algorithms/audioproblems/startstopcut.cpp
```diff
@@ -0,0 +1,118 @@
...
+    "combination of RealAccumulator on the streaming mode.\n"
```
dbogdanov: "In the streaming mode, use it in combination with RealAccumulator"

src/algorithms/audioproblems/startstopcut.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  Real start, stop;
+  uint sFrame, eFrame;
+
+  // looks for the first non-silent frame
+  findNonSilentFrame(audio, sFrame);
+
+  
+  std::vector<Real> reversedAudio = audio;
+  std::reverse(reversedAudio.begin(), reversedAudio.end());
```
dbogdanov: Could this be optimized? No need to reverse the entire audio, as we are only interested in its end.

src/algorithms/audioproblems/startstopcut.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  // sets the start/stop flags according to the thresholds
+  start = (Real)(_hopSize * sFrame) / _sampleRate;
+  stop = (Real)(_hopSize * eFrame) / _sampleRate;
+
+  (start < _maximumStartTime) ? startCut = true : startCut = false;
+  (stop < _maximumStopTime) ? stopCut = true : stopCut = false;
+}
+
+void StartStopCut::findNonSilentFrame(std::vector<Real> audio, uint &nonSilentFrame) {
```
dbogdanov: This can be further optimized depending on what we want. To predict ```startCut```, we do not necessarily need to locate the frame once it is clear that its location would be after the time threshold. Is it still useful to have start and stop positions?

test/src/unittests/audioproblems/test_startstopcut.py
```diff
@@ -0,0 +1,95 @@
...
+from essentia_test import *
+from essentia import array as esarr
+
+
+class TestStartStopCut(TestCase):
+
+    def InitStartStopCut(self, **kwargs):
+        return StartStopCut(**kwargs)
+
+    def testZero(self):
```
dbogdanov: Add test on a long silence?

Issue #529, opened 2016-12-12
Implement generic Histogram algorithm


dbogdanov:
Implement ```Histogram``` streaming algorithm that maps an input stream of Real values into a configurable histogram. 

Parameters:
- normalize {none; unit_sum,unit_max}; none by default
- minValue
- maxValue
- numberBins (equal-width bins)

dbogdanov:
Implemented in PR #796.

PR #698, opened 2018-03-26
Generic histogram - sample

albincorreya:
#529 
Computes a generic histogram from an input array of real values. The program returns an array of array with a binRange and the frequency of items in those ranges.

[NOTE : have to fix some type conversion errors.] 

[TODO]
* Fix some known bugs.
* write some basic streaming extractors
* unit tests
* add functions to compute probability density functions (if required) ?


dbogdanov:
Superseeded by #796

PR #702, opened 2018-03-31
chgs for gaia2 and fixes for essentia - Ubuntu

sildeag:
To allow Gaia built with the `--with-gaia-qt5` option to be used by Essentia.  The revised Gaia now supports Python2 or Python3.  The included file location for Eigen3 is changed in Gaia to 3rdparty folder and the Eigen version is 3.3.4 from Github.  At present this may include too much and may need to be pared back and `./waf` build instruction on how to do this are welcome.  Cyclops doesn't work under Qt5 due to deprecated QHttp but this has no impact on Essentia.


sildeag:
The only problem with checking Essentia integration with Gaia is that `waf` has to be correctly built for `python2` or `python3` to include both `qt4`, which is now in `waflib` extras as it is almost obsolete, and `qt5`.  Then the `libgaia2.a` used in creating the `essentia.so` must be consistent, ie. qt4 or qt5, with the `--with-gaia` or `--with-gaia-qt5` for the link step.  Plus, I only have `Ubuntu` as a test platform.

sildeag:
Conflict introduced by change to `src/python/wscript` master has been eliminated.  It made no difference in Ubuntu environment.

sildeag:
In my testing:
`python3 ./waf configure --with-python --with-gaia --with-gaia-qt5 --with-examples --with-vamp --with-cpptests` generated all expected programs and libraries.

`python3 ./waf configure --with-python --with-gaia --with-gaia-qt5 --with-example=streaming_extractor_music_svm --with-vamp --with-cpptests`.  Ignored generation of python library and the final cpptests step.  Not sure why it did this - probably something simple - but hopefully this is better than it was.

sildeag:
In the 2 "conflicting files":  `src/python/wscript` requires the changes made since `libpath = '..'` allows the module to link correctly with the shared module `essentia.so` found in the folder up one level.  Without the changed the module used might be the previously installed version.  This would also help fix a docker env where there is no previously installed version of `essentia`.  `src/wscript` requires the changes made to generate the `essentia.pc` file, and fix issues with building `examples` and `vamp` on Ubuntu.  This previously worked in older versions of MTG `essentia` but does not function currently in the MTG version for Ubuntu.

sildeag:
`travis/build_wheels.sh` uses 2 arguments which default to the functionality MTG wants.  However, these 2 arguments allow testing of `build_wheels.sh`.  The `build_wheels.sh` change can be ignored, if it is a problem.

sildeag:
Added `--only-python` option to bring scripts files up-to-date with MTG.  Closer look will reveal "No conflicts."

sildeag:
Dimtry, at the moment `--only-python` does not work as it supposed to.  I am revising the build files so that it does - it will take a day or two to fix.  As an enhancement, I could integrate the process so that `only-python` triggers a build `--build-static --static-dependencies` but I'll leave that alone.

sildeag:
Still needing to resolve building `--with-examples`.

sildeag:
Conflict with `build_wheels.sh` resolved (1 line spacing difference).  Still outstanding issue with linking of Gaia example.

dbogdanov:
Closing, as Gaia migrated to Qt5 in https://github.com/MTG/gaia/pull/92 for the next release.

Issue #718, opened 2018-05-02
Add chromaprint to MusicExtractor and implement a standalone chromaprint extractor


dbogdanov:
1) Add an option in  [MusicExtractor](http://essentia.upf.edu/documentation/reference/std_MusicExtractor.html) to compute chromaprints using the Chromaprinter algorithm. This should be configurable via a parameter or the [profile file](http://essentia.upf.edu/documentation/streaming_extractor_music.html#configuration) and disabled by default.

2) Implement a standalone executable chromaprint extractor.

PR #720, opened 2018-05-07
MusicExtractor: compute a chromaprint when libchromaprint is available

pabloEntropia:
This pull request modifies [MusicExtractor](http://essentia.upf.edu/documentation/reference/std_MusicExtractor.html) to compute a Chromaprint  when libchromaprint is installed as demanded on  #718.
In such case two new parameters are added to MusicExtractor:

- **chromaprintCompute** : a flag to compute the Chromaprint. Defaults to `false`.
- **chromaprintDuration** : the amount of time from the beginning used to compute the Chromaprint. 0 to use the full audio length. Defaults to the full audio length.

The parameters can also be configured through the Yaml file with the following lines as explained [here](http://essentia.upf.edu/documentation/streaming_extractor_music.html#configuration):

```
chromaprint:
<tab>duration: 10.0
<tab>compute: true
```

### DIFF
--- a/src/algorithms/extractor/musicextractor.cpp
+++ b/src/algorithms/extractor/musicextractor.cpp
@@ -84,6 +84,11 @@ void MusicExtractor::configure() {
   rhythmStats = parameter("rhythmStats").toVectorString();
   mfccStats = parameter("mfccStats").toVectorString();
   gfccStats = parameter("gfccStats").toVectorString();
+  
+#if HAVE_LIBCHROMAPRINT
+  chromaprintCompute = parameter("chromaprintCompute").toBool();
+  chromaprintDuration = parameter("chromaprintDuration").toReal();
+#endif
 
 #if HAVE_GAIA2 
   if (parameter("highlevel").isConfigured()) { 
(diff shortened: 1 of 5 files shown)


src/algorithms/extractor/musicextractor.cpp
```diff
@@ -585,5 +610,42 @@ void MusicExtractor::setExtractorOptions(const std::string& filename) {
...
+    chromaprinter->compute();
+
+    results.add("chromaprint.string", chromaprint);
+    if (chromaprintDuration == 0.f)
+      results.add("chromaprint.duration", results.value<Real>("metadata.audio_properties.length"));
+    else
+    results.add("chromaprint.duration", chromaprintDuration);
+  }
+  catch (const EssentiaException&) {
+    throw EssentiaException("File looks like a completely silent file");
```
dbogdanov: This is not informative. Use the text from the original exception (thrown by MonoLoader or Chromaprinter) instead.

dbogdanov:
It makes sense to add the new yaml profile options for chromaprinter to the [documentation page for the standalone extractor](https://github.com/MTG/essentia/blob/master/doc/sphinxdoc/streaming_extractor_music.rst) before we merge this.

Issue #722, opened 2018-05-09
MusicExtractor throws exception on very high ReplayGain values (silent files)


dbogdanov:
When MusicExtractor computes replay_gain value, it checks it for being unusually high, which would be the case for silent files, and an exception is thrown. 

File example: https://freesound.org/people/maxiferado/sounds/363613/

We need a more clear strategy of what to do. Perhaps sending an E_WARNING instead of throwing exception is better.

PR #735, opened 2018-05-25
Remove error thrown when replayGain is higher than threshold

lorenzo-romanelli:
When using the musicextractor algorithm, an error is thrown when the computed replayGain is higher than a threshold (40dB), and the extraction fails (see #722).
Such behaviour is not ideal (a warning should be thrown instead), so here it was removed.


dbogdanov:
I see a modification to Music Extractor, but not to Freesound Extractor. Is that right?

lorenzo-romanelli:
Yes I only worked on the music extractor (freesound extractor doesn't calculate replayGain).

dbogdanov:
We may still prefer to have an error in the context of some applications. We should also think about replacing loudness normalization from replayGain to a common EBU R128 LUFS. Therefore, closing this for now.

PR #742, opened 2018-06-05
Minimum peak distance parameter for PeakDetection

pabloEntropia:
This PR adds a new parameter `minPeakDistance` that sets the minimum distance between contiguous peaks. A value of `0` (by default) will bypass this feature. The algorithm uses an amplitude criterium to choose which peaks to delete.
peaks

### DIFF
--- a/src/algorithms/standard/peakdetection.cpp
+++ b/src/algorithms/standard/peakdetection.cpp
@@ -44,6 +44,7 @@ void PeakDetection::configure() {
   _range = parameter("range").toReal();
   _interpolate = parameter("interpolate").toBool();
   _orderBy = parameter("orderBy").toLower();
+  _minPeakDistance = parameter("minPeakDistance").toReal();
 
   if (_minPos >= _maxPos) {
     throw EssentiaException("PeakDetection: The minimum position has to be less than the maximum position");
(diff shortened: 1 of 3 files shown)


src/algorithms/standard/peakdetection.cpp
```diff
@@ -165,6 +166,43 @@ void PeakDetection::compute() {
...
+      
+      for (int l = 0; l < deletedPeaks.size(); l++)
+        peaks.erase(peaks.begin() + deletedPeaks[l]);
+        
+      deletedPeaks.clear();
+      deletedPeaks.reserve(peaks.size());
+      k++;
+    }
+
+    std::sort(peaks.begin(), peaks.end(),
```
dbogdanov: In the case of ```_orderBy == "amplitude"``` and ```_minPeakDistance > 0``` the peaks are first sorted by magnitude, then sorted by position and then sorted by magnitude again. This can be optimized.

src/algorithms/standard/peakdetection.h
```diff
@@ -55,6 +56,7 @@ class PeakDetection : public Algorithm {
     declareParameter("threshold", "peaks below this given threshold are not output", "(-inf,inf)", -1e6);
     declareParameter("orderBy", "the ordering type of the output peaks (ascending by position or descending by value)", "{position,amplitude}", "position");
     declareParameter("interpolate", "boolean flag to enable interpolation", "{true,false}", true);
+    declareParameter("minPeakDistance", "minimum distance between consecutive peaks (0 to bypass this feature)", "[0,inf)", 0.0);
```
dbogdanov: Maybe it's worth to mention in the doc (here or in the description DOC string, that minPeakDistance relies on picking up the predominant peaks)

dbogdanov:
@pabloEntropia Updates?

PR #774, opened 2018-08-02
Use const char* for compatibility with Python 3.7

ffont:
Minor changes so that Essentia compiles using Python 3.7.
Tested on macOS only.

### DIFF
--- a/src/python/pystreamingalgorithm.cpp
+++ b/src/python/pystreamingalgorithm.cpp
@@ -166,7 +166,7 @@ PyObject* PyStreamingAlgorithm::configure (PyStreamingAlgorithm* self, PyObject*
 }
 
 PyObject* PyStreamingAlgorithm::hasSink(PyStreamingAlgorithm* self, PyObject* obj) {
-  char* name = PyString_AsString(obj);
+  const char* name = PyString_AsString(obj);
   if (name == NULL) {
     PyErr_SetString(PyExc_ValueError, "Algorithm.hasSink requires 1 string argument");
     return NULL;
@@ -177,7 +177,7 @@ PyObject* PyStreamingAlgorithm::hasSink(PyStreamingAlgorithm* self, PyObject* ob
 }
 
 PyObject* PyStreamingAlgorithm::hasSource(PyStreamingAlgorithm* self, PyObject* obj) {
-  char* name = PyString_AsString(obj);
+  const char* name = PyString_AsString(obj);
   if (name == NULL) {
     PyErr_SetString(PyExc_ValueError, "Algorithm.hasSource requires 1 string argument");
     return NULL;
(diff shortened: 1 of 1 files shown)


sildeag:
I don't think allowing Python 3.7 support for Macs is reasonable just because `wscript` causes no conflicts ie. requires no change to `wscript` and therefore presents no conflicts with the base branch.   Then you get a patchwork of some OS supporting Python 2/3 and others not.   Ubuntu OS requires `pyext_PATTERN` to be fixed so that its `import` works too.  That might create a conflict with the base branch but it would allow Ubuntu OS to work with python 3.6 or 3.7 and be helpful to other users.

sildeag:
@dbogdanov I'll attempt to clarify.  With `python3` on Ubuntu code like this is required to be consistent with how other peoples python packages are installed.  
   `if sys.platform.startswith('linux'):`
        ` ``conf.env['pyext_PATTERN'] = "%s.so"`
        ` ``python_dir = str(conf.env['PYTHONDIR'])`
        ` ``python_dir = python_dir.replace ('3/', str(conf.env['PYTHON_VERSION'])+'/', 1)`
        ` ``conf.env['PYTHONDIR'] = python_dir`

Obviously, since you haven't merged any of my changes for either `Essentia` or `Gaia` nobody else uses `python3` on `Ubuntu` or they manually move the python distribution, set their python path differently, etc..  I don't really know.  Any merges increase the chances of conflict from changes made to the previous `master` on already tested code, from code that came into existence later.  Risks obsoleting functionality before it has even a chance of being tested and used by others.  I had included these particular code changes in my code changes, these aren't really problematic, but other code changes might be.  I don't recall seeing script to change `pyext_PATTERN` for linux OS but it might have slip in.

dbogdanov:
This looks hacky. For PYTHONDIR we should rely on waf as much as we can, and it uses [distutil](https://gitlab.com/ita1024/waf/blob/master/waflib/Tools/python.py#L492).

The problem is that for the default installation prefix to ```/usr/local```  the output path ```/usr/local/lib/python3/dist-packages``` is not included in the default python ```sys.path```, but the ``/usr/lib/python3/dist-packages`` path is.

```
$python3 -c "from distutils.sysconfig import get_python_lib; print(get_python_lib(standard_lib=0, prefix='/usr/local'))"
/usr/local/lib/python3/dist-packages

$python3 -c "from distutils.sysconfig import get_python_lib; print(get_python_lib(standard_lib=0))"
/usr/lib/python3/dist-packages
```

We stated in the documentation, that the user should make sure that path is included. It is also possible to specify PYTHONDIR with the ```--pythondir``` flag.

Here's a [similar issue with discussion](https://github.com/pothosware/SoapySDR/issues/129).

sildeag:
It is a hack, `for Linux OSs`, but better to have a feature that works and puts the package where every other python3.5, python 3.6, python 3.7, etc. goes than have it placed in `python3/dist-packages`.  I've noticed no other python3 package seems to have this issue on Ubuntu using default installation.

PR #791, opened 2018-10-03
CrossSimilarityMatrix and CoverSongSimilarity algo implementations

albincorreya:
This PR adds the following algos and some math utility functions in essentia.

### CrossSimilarityMatrix

- Computes a cross similarity matrix from two 2D input feature vectors 
  - Option to binarize the Euclidean similarity matrix

### ChromaCrossSimilarity
-Computes a binary cross similarity matrix from two input chroma features 
   - Key Invariance using optimal transposition index (OTI).
  - Option to compute the binary similarity matrix using oti-based similarity.

### CoverSongSimilarity

- Computes [Smith-Waterman](https://en.wikipedia.org/wiki/Smith%E2%80%93Waterman_algorithm) local alignment score matrix from an input 2D cross similarity matrix using various specific alignment constraints for the task of cover song identification.

### essentiamath.h

- ```percentile```: qth percentile of a 1D array.
- ```pairwiseDistance```:  pairwise Euclidean distance of two 2D vectors.
- ```heavisideStepFunction```: [Step function](https://en.wikipedia.org/wiki/Heaviside_step_function) for binarizing  an input 2D vector. 
- ```rotateChroma```: rotate chroma bin in the input chromagram by an given index. 
- ```sumFrames```: sum the sum of frames of an input array

### examples
- add a cpp example of standard mode crosssimilaritymatrix.cpp
- add a standard essentia cover song similarity extractor. It outputs cover song similarity score matrix to a json file from given query and reference audio file inputs. 
- added a streaming mode example of crosssimilaritymatrix.cpp
- added a streaming mode example of coversongsimilarity.cpp

### Fixes

- Added standard stl VECTOR_VECTOR_REAL support in essentia python wrapper.
- Henceforth, 2D arrays can be used as a parameter to an algorithm within python interface.


src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+        // pairwise euclidean distance
+        pdistances = pairwiseDistance(queryFeature, referenceFeature);
+
+        // free memorry of input feature arrays
+        std::vector<std::vector<Real> >().swap(queryFeature);
+        std::vector<std::vector<Real> >().swap(referenceFeature);
+    }
+
+    if (pdistances.empty())
+       throw EssentiaException("empty array found while calculating euclidian similarity");
```
dbogdanov: More detailed text: "CrossSimilarityMatrix: empty array found ... "

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    // transposing the array of pairwsie distance
+    std::vector<std::vector<Real> > tpDistances = transpose(pdistances);
+
+    std::vector<std::vector<Real> > ephX(pdistances.size());
+    std::vector<std::vector<Real> > ephY(tpDistances.size());
+
+    std::vector<Real> tempXrow;
```
dbogdanov: Using tempXrow appears redundant as it's size is always 1. Populate ``ephX`` directly?

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+    for (size_t i=0; i<pdistances.size(); i++) {
+        tempXrow.push_back(percentile(pdistances[i], _kappa));
+        ephX[i] = tempXrow;
+        tempXrow.clear();
+    }
+
+    std::vector<Real> tempYrow;
```
dbogdanov: Same here

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+    /*
+    if (similarityX.empty() == 0) {
+      throw EssentiaException("Sim X is empty");
+    }
+
+    if (similarityY.empty() == 0) {
+      throw EssentiaException("Sim Y is empty");
+    }
+    */
+    // Clear memory
```
dbogdanov: Destructors should take care of that.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+    std::vector<std::vector<Real> >().swap(ephX);
+    std::vector<std::vector<Real> >().swap(ephY);
+
+    // Binarise the array with heavisideStepFunction
+    heavisideStepFunction(similarityX);
+    heavisideStepFunction(similarityY);
+
+    csm.assign(similarityX.size(), std::vector<Real>(similarityY.size(), 0));
+    // tranpose similarityY vector for dot product
+    std::vector<std::vector<Real> > tSimilarityY = transpose(similarityY);
```
dbogdanov: Can rewrite ```similarityY``` with its transposed version.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+    // Finally we construct out cross similarity matrix by doing dot product
+    for (size_t x=0; x<similarityX.size(); x++) {
+        for (size_t y=0; y<tSimilarityY.size(); y++) {
+            csm[x][y] = dotProduct(similarityX[x], tSimilarityY[y]);
+        }
+    }
+}
+
+
+// Construct a stacked chroma embedding from an input chroma audio feature vector [TODO] future use beat-synchronised stacked embeddings
```
dbogdanov: Move the todo comment to next line. Format for todos: ``// TODO stuff to do``

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+    return timeEmbedding;
+}
+
+
+// computes global averaged chroma hpcp as described in [1]
+std::vector<Real> CrossSimilarityMatrix::globalAverageChroma(std::vector<std::vector<Real> >& inputFeature) const {
+
+    std::vector<Real> globalChroma;
+    std::vector<std::vector<Real> > featureCopy;
+    // tranpose the array from (time_axis, chroma_bin) to (chroma_bin, time-axis)for easy calculation
```
dbogdanov: To avoid memory allocation for the copy, we can compute the sum value directly on the (time_axis, chroma_bin) matrix.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+    featureCopy = transpose(inputFeature);
+
+    // sum all the value along the time axis of chroma hpcp feature
+    for (size_t i=0; i<featureCopy.size(); i++) {
+        globalChroma.push_back(std::accumulate(featureCopy[i].begin(), featureCopy[i].end(), 0));
+    }
+
+    double maxItem = *std::max_element(globalChroma.begin(), globalChroma.end());
```
dbogdanov: Use ```normalize``` funcion from essentiamath.h

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+    // divide the sum array by the max element to normalise it to 0-1 range
+    for (size_t j=0; j<globalChroma.size(); j++) {
+        globalChroma[j] = globalChroma[j] / maxItem;
+    }
+    return globalChroma;
+}
+
+
+// Compute the optimal transposition index for transposing reference song feature to the musical key of query song feature as described in [1].
+int CrossSimilarityMatrix::optimalTranspositionIndex(std::vector<std::vector<Real> >& featureA, std::vector<std::vector<Real> >& featureB, int nshifts) const {
```
dbogdanov: It makes sense to rename ``featureA`` and ``featureB`` to ``chromaA`` and ``chromaB`` for consistency with other variable names.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    std::vector<Real> globalChromaA = globalAverageChroma(featureA);
+    std::vector<Real> globalChromaB = globalAverageChroma(featureB);
+    std::vector<Real> chromaBcopy = globalChromaB;
```
dbogdanov: No need to store the copy?

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,233 @@
...
+    for(int i=1; i<=nshifts; i++) {
+        // circular rotate the input globalchroma by an index 'i'
+        std::rotate(globalChromaB.begin(), globalChromaB.end() - i, globalChromaB.end());
+        //rotateByIndex(globalChromaB, i);
+        // compute the dot product of the query global chroma and the shifted global chroma of reference song and append to an array
+        valueAtShifts.push_back(dotProduct(globalChromaA, globalChromaB));
+        globalChromaB = chromaBcopy;
+    }
+    // compute the optimal index by finding the index of maximum element in the array of value at various shifts
+    int maxValue = argmax(valueAtShifts);
```
dbogdanov: One-liners are fine: ```return argmax(...);```

src/essentia/essentiamath.h
```diff
@@ -882,6 +882,111 @@ TNT::Array2D<T> transpose(const TNT::Array2D<T>& m) {
   return result;
 }
 
+
+/**
+ * returns the dot product of two 1-D vectors.
+ */
+template <typename T>
+T dotProduct(const std::vector<T>& xArray, const std::vector<T>& yArray) {
+  if (xArray.empty())
```
dbogdanov: ``xArray.empty() || yArray.empty()``

src/essentia/essentiamath.h
```diff
@@ -882,6 +882,111 @@ TNT::Array2D<T> transpose(const TNT::Array2D<T>& m) {
...
+  std::rotate(inputArray.begin(), inputArray.end() - indx, inputArray.end());
+}
+
+
+// returns the n-th percentile of an input array
+template <typename T> T percentile(const std::vector<T>& array, int npercentile) {
+  if (array.empty())
+    throw EssentiaException("trying to calculate percentile of empty array");
+
+  // median has sense only on sorted array
```
dbogdanov: A redundant comment, remove.

src/essentia/essentiamath.h
```diff
@@ -882,6 +882,111 @@ TNT::Array2D<T> transpose(const TNT::Array2D<T>& m) {
...
+  std::vector<T> sorted_array = array;
+  std::sort(sorted_array.begin(), sorted_array.end());
+
+  uint size = sorted_array.size()+1;
+  float nth = npercentile/100.;
+  int idx = size*nth;
```
dbogdanov: Is truncation intentional? We should check the specs whether to truncate or round to compute the percentiles correctly (check how it's done in scipy).

src/essentia/essentiamath.h
```diff
@@ -882,6 +882,111 @@ TNT::Array2D<T> transpose(const TNT::Array2D<T>& m) {
...
+template <typename T> T percentile(const std::vector<T>& array, int npercentile) {
+  if (array.empty())
+    throw EssentiaException("trying to calculate percentile of empty array");
+
+  // median has sense only on sorted array
+  std::vector<T> sorted_array = array;
+  std::sort(sorted_array.begin(), sorted_array.end());
+
+  uint size = sorted_array.size()+1;
+  float nth = npercentile/100.;
```
dbogdanov: Use Real

src/essentia/essentiamath.h
dbogdanov: Merge lines 920 and 921 in one line.

src/essentia/essentiamath.h
```diff
@@ -882,6 +882,111 @@ TNT::Array2D<T> transpose(const TNT::Array2D<T>& m) {
...
+ * Apply heaviside step function to an input m X n dimentional vector.
+ * f(x) = if x<0: x=0; if x>=0: x=1
+ * returns a 2d binary vector of m X n shape
+ */
+template <typename T>
+void heavisideStepFunction(std::vector<std::vector<T> >& inputArray) {
+  if (inputArray.empty())
+    throw EssentiaException("empty array as input");
+
+  //std::vector<std::vector<T> > result(inputArray.size(), std::vector<T>(inputArray[0].size(), 0));
```
dbogdanov: Remove unnecessary comments

src/essentia/essentiamath.h
```diff
@@ -882,6 +882,111 @@ TNT::Array2D<T> transpose(const TNT::Array2D<T>& m) {
...
+        inputArray[i][j] = 0;
+        //result[i][j] = 0;
+      }
+      else if (inputArray[i][j] >= 0) {
+        inputArray[i][j] = 1;
+        //result[i][j] = 1;
+      }
+    }
+  }
+  //return result;
```
dbogdanov: ``return;``

src/essentia/essentiamath.h
```diff
@@ -882,6 +882,111 @@ TNT::Array2D<T> transpose(const TNT::Array2D<T>& m) {
...
+ * Returns a m X k dimentional vector
+ * TODO: [add other distance metrics beside euclidean such as cosine, mahanalobis etc]
+ */
+template <typename T>
+std::vector<std::vector<T> > pairwiseDistance(const std::vector<std::vector<T> >& m, const std::vector<std::vector<T> >& n) {
+
+  //size_t nrows = m.size();
+  size_t xNcols = m.size();
+  size_t yNcols = n.size();
+  double item = 0;
```
dbogdanov: Using double for better precision?

src/essentia/essentiamath.h
```diff
@@ -882,6 +882,111 @@ TNT::Array2D<T> transpose(const TNT::Array2D<T>& m) {
...
+  std::vector<std::vector<T> > result(xNcols, std::vector<T>(yNcols, 0));
+
+  /*
```
dbogdanov: Remove unnecessary commented code

src/examples/standard_crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,64 @@
...
+  // audioloader -> framecutter -> windowing -> FFT -> MFCC
+
+  AlgorithmFactory& factory = standard::AlgorithmFactory::instance();
+
+  Algorithm* csm = factory.create("CrossSimilarityMatrix");
+
+  /////////// CONNECTING THE ALGORITHMS ////////////////
+  cout << "-------- connecting algos ---------" << endl;
+
+  vector<vector<Real> > queryFeatureinput(1000, vector<Real>(12, 1));
```
dbogdanov: It's better to have a real example instead of dummy values. We should have an extractor that output a similarity score given two input files.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    TNT::Array2D<Real> simX = vecvecToArray2D(similarityX);
+    TNT::Array2D<Real> simY = vecvecToArray2D(tSimilarityY);
+    TNT::Array2D<Real> csmOut = TNT::operator*(simX, simY);
+    csmOutput[0] = csmOut;
+    releaseData();
+  }
+  // should not arrive here
+  else EXEC_DEBUG("CrossSimilarityMatrix: wrong type parameter 'otiBinary'. Expect boolean type");
```
dbogdanov: Again, this should be ok to remove, because of ```toBool()```.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,438 @@
...
+  for (size_t i=0; i<chromaA.size(); i++) {
+    for (size_t j=0; j<chromaB.size(); j++) {
+      // compute OTI-based similarity for each frame of chromaA and chromaB
+      for(int k=0; k<=nshifts; k++) {
+        chromaBcopy = chromaB[j];
+        std::rotate(chromaBcopy.begin(), chromaBcopy.end() - k, chromaBcopy.end());
+        valueAtShifts.push_back(dotProduct(chromaA[i], chromaBcopy));
+        chromaBcopy = chromaB[j];
+      }
+      otiIndex = argmax(valueAtShifts);
```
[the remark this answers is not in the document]
albincorreya: yes, the value of ```otiIndex``` is estimated locally for each pair and then you binarize each element of the output similarity matrix according to whether it's less or greater than a semitone.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,438 @@
...
+    }
+    return timeEmbedding;
+  }
+}
+
+// Computes a binary similarity matrix from two chroma vector inputs using OTI as mentioned in [3]
+std::vector<std::vector<Real> > chromaBinarySimMatrix(std::vector<std::vector<Real> >& chromaA, std::vector<std::vector<Real> >& chromaB, int nshifts, Real matchCoef, Real mismatchCoef) {
+
+  int otiIndex;
+  std::vector<Real> valueAtShifts;
```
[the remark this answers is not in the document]
albincorreya: I think ```valueAtShifts``` is apter here since we are computing the dot product at each shift and storing to its respective index.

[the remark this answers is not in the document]
dbogdanov: ok!

src/algorithms/highlevel/crosssimilaritymatrix.h
```diff
@@ -0,0 +1,129 @@
...
+    declareInput(_queryFeature, "queryFeature", "input chromagram of the query song");
+    declareInput(_referenceFeature, "referenceFeature", "input chromagram of the reference song");
+    declareOutput(_csm, "csm", "2D binary cross-similarity matrix of the query and reference features");
+   }
+
+   void declareParameters() {
+    declareParameter("tau", "time delay for embedding in units of number of windows", "[1,inf)", 1);
+    declareParameter("embedDimension", "embedding dimension for the stacked feature embedding. Choose 'embedDimension=1' to use raw input feature vector for the similarity calculation.", "[0,inf)", 9);
+    declareParameter("kappa", "fraction of mutual nearest neighbours to consider while computing euclidean distances", "[0,1]", 0.095);
+    declareParameter("oti", "whether to transpose the key of the reference song to the query song by (OTI)", "{true,false}", true);
```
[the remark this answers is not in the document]
albincorreya: I'd refactored it into one generic ```CrossSimilarityMatrix``` and the other one to be a specific ```ChromaCrossSimilarity``` algorithm.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+
+} // namespace streaming
+} // namespace essentia
+
+
+// computes global averaged chroma hpcp as described in [1]
+std::vector<Real> globalAverageChroma(std::vector<std::vector<Real> >& inputFeature) {
+
+  size_t numbins = inputFeature[0].size();
```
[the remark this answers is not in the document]
albincorreya: added a ```sumFrames``` function in ```essentiamath.h``` which returns the sum of frames over a specified axis

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,438 @@
...
+    if (!shouldStop()) return status;
+
+    // if shouldStop is true, that means there is no more audio coming, so we need
+    // to take what's left to fill in half-frames, instead of waiting for more
+    // data to come in (which would have done by returning from this function)
+
+    int available = input("queryFeature").available();
+    if (available == 0) return NO_INPUT;
+
+    input("queryFeature").setAcquireSize(available);
```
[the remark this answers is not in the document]
albincorreya: in this case, we pad the array until we fill the ```_minFrameSize``` criteria
https://github.com/MTG/essentia/blob/f2a3c709666d9a8ad3abb483d963135eee9be857/src/algorithms/highlevel/crosssimilaritymatrix.cpp#L226

src/algorithms/highlevel/crosssimilaritymatrix.cpp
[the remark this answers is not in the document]
dbogdanov: Not sure about the option to specify the axis. This may be misleading because: ```sumFrames``` mean sum values across frames, not across features (just like other *Frames methods do). Do you actually use this option?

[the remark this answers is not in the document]
albincorreya: Okay, I just tried to make it similar to numpy funcs are. Anyway, I thought it might be useful for some use-cases. 
Yes, I only need to sum the frames. So, just removed the parameter to specify the axis as you suggested.

src/algorithms/highlevel/crosssimilaritymatrix.h
```diff
@@ -0,0 +1,129 @@
...
+   CrossSimilarityMatrix() {
+    declareInput(_queryFeature, "queryFeature", "input chromagram of the query song");
+    declareInput(_referenceFeature, "referenceFeature", "input chromagram of the reference song");
+    declareOutput(_csm, "csm", "2D binary cross-similarity matrix of the query and reference features");
+   }
+
+   void declareParameters() {
+    declareParameter("tau", "time delay for embedding in units of number of windows", "[1,inf)", 1);
+    declareParameter("embedDimension", "embedding dimension for the stacked feature embedding. Choose 'embedDimension=1' to use raw input feature vector for the similarity calculation.", "[0,inf)", 9);
+    declareParameter("kappa", "fraction of mutual nearest neighbours to consider while computing euclidean distances", "[0,1]", 0.095);
```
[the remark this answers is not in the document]
albincorreya: added a parameter ```toBinary``` in the generic cross-similarity algorithm. ```toBinary``` sounds better than ```binarize``` to me. What do you think?

src/algorithms/highlevel/crosssimilaritymatrix.cpp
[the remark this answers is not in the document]
albincorreya: Maybe I wasn't clear with the explanation. ```valueAtShifts``` stores the dotProduct of each row of chromaA and chromaB at every 12 shifts. So I am resetting ```valueAtShifts``` after the inner loop corresponds to 12 shifts get finished. (see line:427)

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,438 @@
...
+    }
+  }
+  // Use default cross similarity computation method based on euclidean distances
+  else {
+
+    std::vector<std::vector<Real> > pdistances;
+    // check whether to transpose by oti
+    if (_oti == true) {
+      int otiIdx = optimalTranspositionIndex(queryFeature, referenceFeature, _noti);
+      std::rotate(referenceFeature.begin(), referenceFeature.end() - otiIdx, referenceFeature.end());
```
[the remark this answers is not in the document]
albincorreya: Yes, you're right! Fixed the bug and added func ```rotateChroma``` in ```essentiamath.h``` to shift the chroma bins in a chromagram by a specified optimal transposition index.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,438 @@
...
+    if (_optimiseThreshold == true) {
+      // optimise the threshold computation on axis X by iniatilizing it to a matrix of ones
+      similarityX.assign(xRows, std::vector<Real>(xCols, 1));
+    }
+    else if (_optimiseThreshold == false) {
+      similarityX.assign(xRows, std::vector<Real>(xCols, 0));
+      // construct thresholded similarity matrix on axis X
+      for (size_t k=0; k<xRows; k++) {
+        for (size_t l=0; l<xCols; l++) {
+          similarityX[k][l] = percentile(pdistances[k], _kappa*100) - pdistances[k][l];
```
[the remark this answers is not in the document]
albincorreya: Fixed

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,438 @@
...
+      for (size_t j=0; j<yCols; j++) {
+        if (similarityY[i][j] < 0) {
+          tSimilarityY[j][i] = 0;
+        }
+        else if (similarityY[i][j] >= 0) {
+          tSimilarityY[j][i] = 1;
+        }
+      }
+    }
+    // finally we construct out cross similarity matrix by multiplying similarityX and similarityY
```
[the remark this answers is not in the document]
albincorreya: Fixed

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <algorithm>
+#include <functional>
+
+namespace essentia {
+namespace standard {
+
+const char* CrossSimilarityMatrix::name = "CrossSimilarityMatrix";
+const char* CrossSimilarityMatrix::category = "Music Similarity";
+const char* CrossSimilarityMatrix::description = DOC("This algorithm computes a euclidean cross-similarity matrix of two sequences of frame features. Similarity values can be optionally binarized\n\n"
+"Use default parameter values for best results while computing cross-similarity using the binarize method.\n\n"
```
dbogdanov: These parameters are good for cover detection, not the other tasks, so mentions that instead (something like "... the default parameters for binarizing are optimized according to [reference] for cover song identification using chroma features..." ).

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+"The input feature arrays should be in the shape (x, y), where 'x' is the number of frames.\n\n"
```
dbogdanov: --" The input feature arrays are vectors of frames of features in the shape  (n_frames, n_features), where 'n_frames' is the number frames, 'n_features' is the number of frame features."

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+"An exception is also thrown if either one of the input audio feature arrays are empty or if the output similarity matrix is empty.\n\n"
```
dbogdanov: As this is a generic algorithm, avoid saying "audio" for features, because someone might want to use other types of features too.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+"References:\n"
+"[1] Serra, J., Serra, X., & Andrzejak, R. G. (2009). Cross recurrence quantification for cover song identification.New Journal of Physics.\n\n");
```
dbogdanov: Missing an empty space: "identification.New"

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+void CrossSimilarityMatrix::configure() {
+  // configure parameters
+  _frameStackStride = parameter("frameStackStride").toInt();
+  _frameStackSize = parameter("frameStackSize").toInt();
+  _kappa = parameter("kappa").toReal();
+  _toBinary = parameter("toBinary").toBool();
+}
+
+// Construct a 'stacked-frames' feature vector from an input audio feature vector by given 'frameStackSize' and 'frameStackStride'
```
dbogdanov: "'stacked-frames' feature vector", I like that description!

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+  if (frameStackSize == 1) {
+    return frames;
+  }
+  size_t stopIdx;
+  int increment = frameStackSize * frameStackStride;
+  std::vector<std::vector<Real> > stackedFrames;
+  stackedFrames.reserve(frames.size() - increment);
+  std::vector<Real> stack;
+  for (size_t i=0; i<(frames.size() - increment); i+=frameStackStride) {
```
dbogdanov: This looks incorrect. Aren't we supposed to generate a stacked-frame feature for every frame index increment?

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+// Construct a 'stacked-frames' feature vector from an input audio feature vector by given 'frameStackSize' and 'frameStackStride'
+std::vector<std::vector<Real> > CrossSimilarityMatrix::stackFrames(std::vector<std::vector<Real> >& frames, int frameStackSize, int frameStackStride) const {
+
+  if (frameStackSize == 1) {
+    return frames;
+  }
+  size_t stopIdx;
+  int increment = frameStackSize * frameStackStride;
+  std::vector<std::vector<Real> > stackedFrames;
+  stackedFrames.reserve(frames.size() - increment);
```
dbogdanov: You know the vector size for the stacked-frame feature, so use it for reserve too: https://stackoverflow.com/a/18992266

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+  if (referenceFeature.empty())
+    throw EssentiaException("CrossSimilarityMatrix: input referenceFeature array is empty.");
+
+  // construct a new vector by stacking the input features by an specified 'frameStackStride' and 'frameStackSize'
+  std::vector<std::vector<Real> >  queryFeatureStack = stackFrames(queryFeature, _frameStackSize, _frameStackStride);
+  std::vector<std::vector<Real> >  referenceFeatureStack = stackFrames(referenceFeature, _frameStackSize, _frameStackStride);
+
+  std::vector<std::vector<Real> > pdistances;
+  // pairwise euclidean distance
+  pdistances = pairwiseDistance(queryFeatureStack, referenceFeatureStack);
```
dbogdanov: Assign directly to ```csm```

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+  // check whether to binarize the euclidean cross-similarity matrix using the given threshold kappa
+  if (_toBinary == true) {
+    std::vector<std::vector<Real> > tpDistances = transpose(pdistances);
+    size_t queryRows = pdistances.size();
```
dbogdanov: It's hard to know what ```queryRows``` and ```referenceRows``` represent from the code. Let's make it even more code-readable. 

Can we call  ```queryRows``` something like a ```queryFeatureSize```? (same for reference)

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+    size_t referenceRows = pdistances[0].size();
+
+    std::vector<Real> thresholdX(queryRows);
```
dbogdanov: These names are still misleading too. Let's put something more meaningful: ```thresholdQuery```? 
Explain in the comment what it means (e..g "Thresholds for distances to reference frame stacks for each query frame stack" if I understood the meaning of the variable names correctly)

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+    std::vector<Real> thresholdY(referenceRows);
+
+    std::vector<std::vector<Real> > outputSimMatrix;
+    outputSimMatrix.assign(queryRows, std::vector<Real>(referenceRows));
+    // construct the binary output similarity matrix using the thresholds computed along the queryFeature axis
+    for (size_t k=0; k<queryRows; k++) {
+      thresholdX[k] = percentile(pdistances[k], _kappa*100);
+      for (size_t l=0; l<referenceRows; l++) {
+        if ((thresholdX[k] - pdistances[k][l]) > 0) {
```
dbogdanov: Same as ```thresholdX[k] > pdistances[k][l]``` (easier to read). Isn't the output supposed to be 0 in this case (distance value is below the threshold)?

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+          outputSimMatrix[k][l] = 1;
+        }
+        else if ((thresholdX[k] - pdistances[k][l]) <= 0) {
+          outputSimMatrix[k][l] = 0.;
```
dbogdanov: Same here

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+        }
+      }
+
+    }
+    // update the binary output similarity matrix by multiplying with the thresholds computed along the referenceFeature axis
+    for (size_t j=0; j<referenceRows; j++) {
+      thresholdY[j] = percentile(tpDistances[j], _kappa*100);
+      for (size_t i=0; i<queryRows; i++) {
+        if ((thresholdY[j] - pdistances[i][j]) > 0) {
```
dbogdanov: Same here. When the distance is below threshold, we should set it to 0

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+          outputSimMatrix[i][j] *= 1; 
+        }
+        else if ((thresholdY[j] - pdistances[i][j]) <= 0) {
```
dbogdanov: Same here, check if correct

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+        }
+        else if ((thresholdX[k] - pdistances[k][l]) <= 0) {
+          outputSimMatrix[k][l] = 0.;
+        }
+      }
+
+    }
+    // update the binary output similarity matrix by multiplying with the thresholds computed along the referenceFeature axis
+    for (size_t j=0; j<referenceRows; j++) {
+      thresholdY[j] = percentile(tpDistances[j], _kappa*100);
```
dbogdanov: The whole point of having the tpDistances is to be able to compute the percentile with the same ```percentile``` function. This is not a good reason, because waste a lot of memory on that. 

We can reserve a temporary std vector before this loop, and recreate the transposed vector for a particular column that we want to pass to ```percentile``` in this loop.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
```diff
@@ -0,0 +1,136 @@
...
+      for (size_t i=0; i<queryRows; i++) {
+        if ((thresholdY[j] - pdistances[i][j]) > 0) {
+          outputSimMatrix[i][j] *= 1;
```
dbogdanov: Skip or set to zero instead of ```*=1``` and ```*=0```. You don't even need to check for the condition to do ```*=1```, just check for the condition to assign to zero.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
albincorreya: Nope, we only do that when ```frameStackStride=1```. So, in this case, it constructs a stacked-frame feature for every frame.

src/algorithms/highlevel/crosssimilaritymatrix.cpp
albincorreya: Yes, ```line::57``` does it. Isn't that what you meant?

src/algorithms/highlevel/crosssimilaritymatrix.cpp
albincorreya: Yes, fixed!

src/algorithms/highlevel/crosssimilaritymatrix.cpp
albincorreya: The step function clips the distance value to 0 if the difference between the threshold and distance is negative.

src/essentia/essentiamath.h
```diff
@@ -882,6 +900,113 @@ TNT::Array2D<T> transpose(const TNT::Array2D<T>& m) {
...
+
+  for (size_t i=0; i<inputArray.size(); i++) {
+    for (size_t j=0; j<inputArray[i].size(); j++) {
+      // initialize all non negative elements as zero
+      if (inputArray[i][j] < 0) {
+        inputArray[i][j] = 0;
+      }
+      else if (inputArray[i][j] >= 0) {
+        inputArray[i][j] = 1;
+      }
```
dbogdanov:
```
inputArray[i][j] = (inputArray[i][j] < 0) ? 0 : 1;
```

test/src/unittests/highlevel/test_chromacrosssimilarity.py
```diff
@@ -0,0 +1,127 @@
...
+        self.assertAlmostEqualVector(self.expected, result)
+
+    def testInvalidParam(self):
+        self.assertConfigureFails(estd.ChromaCrossSimilarity(), { 'binarizePercentile': -1 })
+        self.assertConfigureFails(estd.ChromaCrossSimilarity(), { 'otiBinary': -1 })
+
+    def testOTIBinaryCompute(self):
+        """Tests standard ChromaCrossSimilarity algo when param 'otiBinary=True'"""
+        # test oti-based binary sim matirx method
+        self.assertComputeFails(estd.ChromaCrossSimilarity(otiBinary=True), [])
```
dbogdanov: There's only a fail test on empty input here, but no OTI regresssion.

src/algorithms/highlevel/chromacrosssimilarity.cpp
```diff
@@ -0,0 +1,392 @@
...
+      }
+      _iterIdx++;
+      // clear the internal states after each compute() method call
+      _queryFeatureStack.clear();
+      _referenceFeatureStack.clear();
+      _pdistances.clear();
+      _thresholdQuery.clear();
+      _thresholdReference.clear();
+    }
+    else if (!_streaming) {
```
dbogdanov:
```
else { \\ no streaming
```

test/src/unittests/highlevel/test_chromacrosssimilarity.py
albincorreya: We don't have any reference implementation to test that against ours. And I did a python implementation before which is too slow to run. Should I use that?

dbogdanov: Yes, we can try to use that.

dbogdanov:
Thanks for adding a python example! Merging this.

PR #796, opened 2018-10-15
Review needed on streaming histogram work in progress

Alia-morsi:


src/algorithms/stats/histogram.cpp
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+#include "histogram.h"
+
+using namespace essentia;
+using namespace standard;
+
+const char* Histogram::name = "Histogram";
+const char* Histogram::category = "Statistics";
+const char* Histogram::description = DOC("This algorithm computes a histogram given the following preconfigured parameters: a normalization setting, a minimum value, a maximum value, and the number of equal sized bins for the histogram. ");
```
dbogdanov: No need to describe the parameters in the DOC string, as they will be picked up for documentation from the ```declareParameter()``` methods.

src/algorithms/stats/histogram.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+void Histogram::configure() {
+    _normalize = parameter("normalize").toString();
```
dbogdanov: Use two spaces for indentation for consistency with other C++ code in Essentia

src/algorithms/stats/histogram.cpp
```diff
@@ -0,0 +1,61 @@
...
+  
+  const std::vector<Real>& array = _array.get();
+  std::vector<Real>& binCounts = _binCounts.get();
+  std::vector<Real>& binRanges = _binRanges.get();
+  
+  binCounts.resize(_numberBins);
+  binRanges.resize(_numberBins);
+
+  if(_maxValue < _minValue)
+    throw EssentiaException("Histogram: maxValue must be > minValue");
```
dbogdanov: What if maxValue == minValue?

src/algorithms/stats/histogram.h
```diff
@@ -0,0 +1,93 @@
...
+
+  std::string  _normalize;
+  Real _minValue;
+  Real _maxValue;
+  int  _numberBins;
+
+ public:
+  Histogram() {
+    declareInput(_array, "array", "the input array");
+    declareOutput(_binCounts, "binCounts", "the values in the equally-spaced bins");
```
dbogdanov: The name ```binCounts``` is not the best one because it does not make sense for the normalized histogram. Perhaps, rename to "histogram"?

src/algorithms/stats/histogram.h
```diff
@@ -0,0 +1,93 @@
...
+    declareOutput(_binRanges, "binRanges", "the ranges of the equally-spaced bins");
+  }
+
+  void declareParameters(){
+    declareParameter("normalize", "the normalization setting.", "{none, unit_sum, unit_max}",  std::string("none"));
+    declareParameter("minValue", "the min value of the histogram", "[0, Inf)", 0.0);
+    declareParameter("maxValue", "the max value of the histogram", "[0, Inf)", 1.0);
+    declareParameter("numberBins", "the number of bins", "(0, Inf)", 1);
```
dbogdanov: A user might not care about the number of bins but still, he might want to plot a histogram. Having only one bin by default won't produce a useful output. [NumPy histogram](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.histogram.html) defaults to 10 bins, perhaps we can take that value.

test/src/unittests/stats/test_histogram.py
```diff
@@ -0,0 +1,53 @@
...
+    self.assertConfigureFails(Histogram(), {'maxValue' : -1})
+    self.assertConfigureFails(Histogram(), {'minValue' : -1})
+    self.assertConfigureFails(Histogram(), {'numberBins' : 0})
+    self.assertConfigureFails(Histogram(), {'numberBins' : -1})
+
+  def testInvalidConfigurationCombination(self):
+    self.assertRaises(EssentiaException, Histogram(maxValue = 0, minValue = 1), (zeros(1000)))
+
+  def testRegression(self):
+   inputArray = readVector(join(filedir(), 'stats/input.txt'))
```
dbogdanov: For simplicity, as we do not have many values, you can hardcode those values in the code instead of loading them from files.

src/algorithms/stats/histogram.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  Real bin_width = (_maxValue - _minValue)/(Real)_numberBins;
+
+  binRanges[0] = _minValue + (Real)(bin_width/2.0);
```
dbogdanov: It may be hard to understand the name ```binRanges```. What is computed in the code are the positions of bin centers. NumPy outputs the bin edges (length(hist)+1). I don't have a strong opinion, but perhaps, having edges is more useful then centers. Especially, in the case if we want to add other scales to the algorithm apart from linear in the future (for example, logarithmic)

src/algorithms/stats/histogram.cpp
```diff
@@ -0,0 +1,61 @@
...
+  for(int i = 1; i < _numberBins; i++) {
+    binRanges[i] = binRanges[i-1] + bin_width;
+  }
+
+  for(int i = 0; i < array.size(); i++){
```
dbogdanov: Say in documentation that "Values outside the range are ignored."

src/algorithms/stats/histogram.cpp
```diff
@@ -0,0 +1,61 @@
...
+  std::vector<Real>& binCounts = _binCounts.get();
+  std::vector<Real>& binRanges = _binRanges.get();
+  
+  binCounts.resize(_numberBins);
+  binRanges.resize(_numberBins);
+
+  if(_maxValue < _minValue)
+    throw EssentiaException("Histogram: maxValue must be > minValue");
+
+  Real bin_width = (_maxValue - _minValue)/(Real)_numberBins;
```
dbogdanov: Compute bin_width inside ```configure()``` method.

src/algorithms/stats/histogram.cpp
dbogdanov: Do this check and through this exception in ```configure()```

src/algorithms/stats/histogram.cpp
dbogdanov: We can actually also precompute bin ranges (or bin edges) in ```configure()``` into a variable, and solely assign the algorithm's output to that variable here.

test/src/unittests/stats/test_histogram.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+from essentia_test import *
+
+testdir = join(filedir(), 'histogram')
+
+class TestHistogram(TestCase):
+
+  def testZero(self):
+    self.assertEqualVector(Histogram()(zeros(1000)), [[1000], [0.5]])
```
dbogdanov: For better readability, set the default parameters you use explicitly in the test code.

test/src/unittests/stats/test_histogram.py
```diff
@@ -0,0 +1,53 @@
...
+    self.assertConfigureFails(Histogram(), {'normalize' : 'y'})
+    self.assertConfigureFails(Histogram(), {'maxValue' : -1})
+    self.assertConfigureFails(Histogram(), {'minValue' : -1})
+    self.assertConfigureFails(Histogram(), {'numberBins' : 0})
+    self.assertConfigureFails(Histogram(), {'numberBins' : -1})
+
+  def testInvalidConfigurationCombination(self):
+    self.assertRaises(EssentiaException, Histogram(maxValue = 0, minValue = 1), (zeros(1000)))
+
+  def testRegression(self):
```
dbogdanov: Add a test for maxValue == minValue

test/src/unittests/stats/histogram/ranges.txt
```diff
@@ -0,0 +1,3 @@
+ 0.050022757620501929 0.14998849366365799 0.24995422970681405 0.34991996574997009 0.44988570179312615 0.54985143783628221 0.64981717387943827 0.74978290992259433 0.84974864596575039 0.94971438200890645
```
dbogdanov: Missing the ```input.txt``` file.

Alia-morsi:
Please find responses/questions to each of your comments inline and in red.
Thanks!

On Tue, Oct 16, 2018 at 4:52 PM Dmitry Bogdanov <notifications@github.com>
wrote:

> *@dbogdanov* requested changes on this pull request.
>
> Great!! See some comments left in the review.
> ------------------------------
>
> In src/algorithms/stats/histogram.cpp
> <https://github.com/MTG/essentia/pull/796#discussion_r225552362>:
>
> > + * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
> + * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
> + * details.
> + *
> + * You should have received a copy of the Affero GNU General Public License
> + * version 3 along with this program.  If not, see http://www.gnu.org/licenses/
> + */
> +
> +#include "histogram.h"
> +
> +using namespace essentia;
> +using namespace standard;
> +
> +const char* Histogram::name = "Histogram";
> +const char* Histogram::category = "Statistics";
> +const char* Histogram::description = DOC("This algorithm computes a histogram given the following preconfigured parameters: a normalization setting, a minimum value, a maximum value, and the number of equal sized bins for the histogram. ");
>
> No need to describe the parameters in the DOC string, as they will be
> picked up for documentation from the declareParameter() methods.
>
 Done

> ------------------------------
>
> In src/algorithms/stats/histogram.cpp
> <https://github.com/MTG/essentia/pull/796#discussion_r225554546>:
>
> > + *
> + * You should have received a copy of the Affero GNU General Public License
> + * version 3 along with this program.  If not, see http://www.gnu.org/licenses/
> + */
> +
> +#include "histogram.h"
> +
> +using namespace essentia;
> +using namespace standard;
> +
> +const char* Histogram::name = "Histogram";
> +const char* Histogram::category = "Statistics";
> +const char* Histogram::description = DOC("This algorithm computes a histogram given the following preconfigured parameters: a normalization setting, a minimum value, a maximum value, and the number of equal sized bins for the histogram. ");
> +
> +void Histogram::configure() {
> +    _normalize = parameter("normalize").toString();
>
> Use two spaces for indentation for consistency with other C++ code in
> Essentia
>
Done. Sorry to have missed that in the first place.

> ------------------------------
>
> In src/algorithms/stats/histogram.cpp
> <https://github.com/MTG/essentia/pull/796#discussion_r225554843>:
>
> > +    _minValue = parameter("minValue").toReal();
> +    _maxValue = parameter("maxValue").toReal();
> +    _numberBins = parameter("numberBins").toInt();
> +}
> +
> +void Histogram::compute() {
> +
> +  const std::vector<Real>& array = _array.get();
> +  std::vector<Real>& binCounts = _binCounts.get();
> +  std::vector<Real>& binRanges = _binRanges.get();
> +
> +  binCounts.resize(_numberBins);
> +  binRanges.resize(_numberBins);
> +
> +  if(_maxValue < _minValue)
> +    throw EssentiaException("Histogram: maxValue must be > minValue");
>
> What if maxValue == minValue?
>

Also my bad, this edge case is very clear and should have been handled.,

Some options are to:
1) throw an exception and disallow it
2) or to allow the scenario if the number of bins is set to 1 (throw an
exception if minValue == maxValue but numberBins set explicitly > 1) and
have the edges array have only 2 entries, both at the same value
3) In numPy, even if you set the max value the same as the min value, The
actual min edge and the max edge are 0.5 above and below that value, and in
that sense, there is no restriction on the number of bins you can have even
in that case.

- For this pull request I followed 2) above, because I don't know if it is
a good idea to diverge from the minValue and maxValue settings given
explicitly by the user without giving off some warning.
However, a problem is that if a user just inputs maxValue == minValue (and
not put anything in the numberBins parameter), this exception will be
thrown as the default will be set to 10.

Let me know what you think about this.


------------------------------
>
> In src/algorithms/stats/histogram.h
> <https://github.com/MTG/essentia/pull/796#discussion_r225556338>:
>
> > +class Histogram : public Algorithm {
> +
> + protected:
> +  Input<std::vector<Real> > _array;
> +  Output<std::vector<Real> > _binCounts;
> +  Output<std::vector<Real> > _binRanges;
> +
> +  std::string  _normalize;
> +  Real _minValue;
> +  Real _maxValue;
> +  int  _numberBins;
> +
> + public:
> +  Histogram() {
> +    declareInput(_array, "array", "the input array");
> +    declareOutput(_binCounts, "binCounts", "the values in the equally-spaced bins");
>
> The name binCounts is not the best one because it does not make sense for
> the normalized histogram. Perhaps, rename to "histogram"?
>
Done

> ------------------------------
>
> In src/algorithms/stats/histogram.h
> <https://github.com/MTG/essentia/pull/796#discussion_r225562087>:
>
> > +  Real _minValue;
> +  Real _maxValue;
> +  int  _numberBins;
> +
> + public:
> +  Histogram() {
> +    declareInput(_array, "array", "the input array");
> +    declareOutput(_binCounts, "binCounts", "the values in the equally-spaced bins");
> +    declareOutput(_binRanges, "binRanges", "the ranges of the equally-spaced bins");
> +  }
> +
> +  void declareParameters(){
> +    declareParameter("normalize", "the normalization setting.", "{none, unit_sum, unit_max}",  std::string("none"));
> +    declareParameter("minValue", "the min value of the histogram", "[0, Inf)", 0.0);
> +    declareParameter("maxValue", "the max value of the histogram", "[0, Inf)", 1.0);
> +    declareParameter("numberBins", "the number of bins", "(0, Inf)", 1);
>
> A user might not care about the number of bins but still, he might want to
> plot a histogram. Having only one bin by default won't produce a useful
> output. NumPy histogram
> <https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.histogram.html>
> defaults to 10 bins, perhaps we can take that value.
>
Done. Good idea! :)

> ------------------------------
>
> In test/src/unittests/stats/test_histogram.py
> <https://github.com/MTG/essentia/pull/796#discussion_r225562751>:
>
> > +
> +  def testZero(self):
> +    self.assertEqualVector(Histogram()(zeros(1000)), [[1000], [0.5]])
> +
> +  def testOutOfRangeConfiguration(self):
> +    self.assertConfigureFails(Histogram(), {'normalize' : 'y'})
> +    self.assertConfigureFails(Histogram(), {'maxValue' : -1})
> +    self.assertConfigureFails(Histogram(), {'minValue' : -1})
> +    self.assertConfigureFails(Histogram(), {'numberBins' : 0})
> +    self.assertConfigureFails(Histogram(), {'numberBins' : -1})
> +
> +  def testInvalidConfigurationCombination(self):
> +    self.assertRaises(EssentiaException, Histogram(maxValue = 0, minValue = 1), (zeros(1000)))
> +
> +  def testRegression(self):
> +   inputArray = readVector(join(filedir(), 'stats/input.txt'))
>
> For simplicity, as we do not have many values, you can hardcode those
> values in the code instead of loading them from files.
>
I'm using the stats/input.txt file that is already part of essentia
(test/src/unittests/stats/stats/input.txt), and this one has ~25000 values.
So I won't hardcode that one.
However, based on your recommendation, I will hardcode the output arrays
and use them for comparison

> ------------------------------
>
> In src/algorithms/stats/histogram.cpp
> <https://github.com/MTG/essentia/pull/796#discussion_r225567483>:
>
> > +
> +void Histogram::compute() {
> +
> +  const std::vector<Real>& array = _array.get();
> +  std::vector<Real>& binCounts = _binCounts.get();
> +  std::vector<Real>& binRanges = _binRanges.get();
> +
> +  binCounts.resize(_numberBins);
> +  binRanges.resize(_numberBins);
> +
> +  if(_maxValue < _minValue)
> +    throw EssentiaException("Histogram: maxValue must be > minValue");
> +
> +  Real bin_width = (_maxValue - _minValue)/(Real)_numberBins;
> +
> +  binRanges[0] = _minValue + (Real)(bin_width/2.0);
>
> It may be hard to understand the name binRanges. What is computed in the
> code are the positions of bin centers. NumPy outputs the bin edges
> (length(hist)+1). I don't have a strong opinion, but perhaps, having edges
> is more useful then centers. Especially, in the case if we want to add
> other scales to the algorithm apart from linear in the future (for example,
> logarithmic)
>
I didn't have a strong opinion either. and I actually had them as edges at
first, but i got the midpoints idea from octave, and i liked the simplicity
of having both the histogram counts and the centers array as parallel
arrays.

But, change done! now it's like numPy

------------------------------
>
> In src/algorithms/stats/histogram.cpp
> <https://github.com/MTG/essentia/pull/796#discussion_r225568166>:
>
> > +  std::vector<Real>& binRanges = _binRanges.get();
> +
> +  binCounts.resize(_numberBins);
> +  binRanges.resize(_numberBins);
> +
> +  if(_maxValue < _minValue)
> +    throw EssentiaException("Histogram: maxValue must be > minValue");
> +
> +  Real bin_width = (_maxValue - _minValue)/(Real)_numberBins;
> +
> +  binRanges[0] = _minValue + (Real)(bin_width/2.0);
> +  for(int i = 1; i < _numberBins; i++) {
> +    binRanges[i] = binRanges[i-1] + bin_width;
> +  }
> +
> +  for(int i = 0; i < array.size(); i++){
>
> Say in documentation that "Values outside the range are ignored."
>
Added to the description in .cpp file

> ------------------------------
>
> In src/algorithms/stats/histogram.cpp
> <https://github.com/MTG/essentia/pull/796#discussion_r225570042>:
>
> > +    _numberBins = parameter("numberBins").toInt();
> +}
> +
> +void Histogram::compute() {
> +
> +  const std::vector<Real>& array = _array.get();
> +  std::vector<Real>& binCounts = _binCounts.get();
> +  std::vector<Real>& binRanges = _binRanges.get();
> +
> +  binCounts.resize(_numberBins);
> +  binRanges.resize(_numberBins);
> +
> +  if(_maxValue < _minValue)
> +    throw EssentiaException("Histogram: maxValue must be > minValue");
> +
> +  Real bin_width = (_maxValue - _minValue)/(Real)_numberBins;
>
> Compute bin_width inside configure() method.
>
Done. (added bin_width as a private class parameter in that case..)

> ------------------------------
>
> In src/algorithms/stats/histogram.cpp
> <https://github.com/MTG/essentia/pull/796#discussion_r225571698>:
>
> > +    _minValue = parameter("minValue").toReal();
> +    _maxValue = parameter("maxValue").toReal();
> +    _numberBins = parameter("numberBins").toInt();
> +}
> +
> +void Histogram::compute() {
> +
> +  const std::vector<Real>& array = _array.get();
> +  std::vector<Real>& binCounts = _binCounts.get();
> +  std::vector<Real>& binRanges = _binRanges.get();
> +
> +  binCounts.resize(_numberBins);
> +  binRanges.resize(_numberBins);
> +
> +  if(_maxValue < _minValue)
> +    throw EssentiaException("Histogram: maxValue must be > minValue");
>
> Do this check and through this exception in configure()
>
Done

> ------------------------------
>
> In src/algorithms/stats/histogram.cpp
> <https://github.com/MTG/essentia/pull/796#discussion_r225572420>:
>
> > +
> +void Histogram::compute() {
> +
> +  const std::vector<Real>& array = _array.get();
> +  std::vector<Real>& binCounts = _binCounts.get();
> +  std::vector<Real>& binRanges = _binRanges.get();
> +
> +  binCounts.resize(_numberBins);
> +  binRanges.resize(_numberBins);
> +
> +  if(_maxValue < _minValue)
> +    throw EssentiaException("Histogram: maxValue must be > minValue");
> +
> +  Real bin_width = (_maxValue - _minValue)/(Real)_numberBins;
> +
> +  binRanges[0] = _minValue + (Real)(bin_width/2.0);
>
> We can actually also precompute bin ranges (or bin edges) in configure()
> into a variable, and solely assign the algorithm's output to that variable
> here.
>
Ok, done.

------------------------------
>
> In test/src/unittests/stats/test_histogram.py
> <https://github.com/MTG/essentia/pull/796#discussion_r225572987>:
>
> > +# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
> +# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
> +# details.
> +#
> +# You should have received a copy of the Affero GNU General Public License
> +# version 3 along with this program. If not, see http://www.gnu.org/licenses/
> +
> +
> +from essentia_test import *
> +
> +testdir = join(filedir(), 'histogram')
> +
> +class TestHistogram(TestCase):
> +
> +  def testZero(self):
> +    self.assertEqualVector(Histogram()(zeros(1000)), [[1000], [0.5]])
>
> For better readability, set the default parameters you use explicitly in
> the test code.
>
Done.

> ------------------------------
>
> In test/src/unittests/stats/test_histogram.py
> <https://github.com/MTG/essentia/pull/796#discussion_r225573715>:
>
> > +class TestHistogram(TestCase):
> +
> +  def testZero(self):
> +    self.assertEqualVector(Histogram()(zeros(1000)), [[1000], [0.5]])
> +
> +  def testOutOfRangeConfiguration(self):
> +    self.assertConfigureFails(Histogram(), {'normalize' : 'y'})
> +    self.assertConfigureFails(Histogram(), {'maxValue' : -1})
> +    self.assertConfigureFails(Histogram(), {'minValue' : -1})
> +    self.assertConfigureFails(Histogram(), {'numberBins' : 0})
> +    self.assertConfigureFails(Histogram(), {'numberBins' : -1})
> +
> +  def testInvalidConfigurationCombination(self):
> +    self.assertRaises(EssentiaException, Histogram(maxValue = 0, minValue = 1), (zeros(1000)))
> +
> +  def testRegression(self):
>
> Add a test for maxValue == minValue
>
Sure, but from this do I get the hint that you prefer it be treated like an
exception case?

> ------------------------------
>
> In test/src/unittests/stats/histogram/ranges.txt
> <https://github.com/MTG/essentia/pull/796#discussion_r225574552>:
>
> > @@ -0,0 +1,3 @@
> + 0.050022757620501929 0.14998849366365799 0.24995422970681405 0.34991996574997009 0.44988570179312615 0.54985143783628221 0.64981717387943827 0.74978290992259433 0.84974864596575039 0.94971438200890645
>
> Missing the input.txt file.
>
the path is test/src/unittests/stats/stats/input.txt

Other Changes/Comments:

- Added the normalization options and tests (in the testRegression function)

- modified zero test to now include the 10 bin default. But i had to use
assertAlmostEqual when comparing the binCenters arrays because some error
gets introduced... Is this ok?

- There's a test case that would be problematic: if histogram is used with
default configuration (so bins = 10, minValue = 0 and maxValue = 1), 10
bins would be created between 0 and 1, and all values < the min and max
value would be disregarded. I feel reluctant overriding the minValue /
maxValue settings of the user because I can't find out if they were
explicitly set or just kept from the defaults..

Thanks for your patience.

Alia-morsi:
- modified zero test to now include the 10 bin default. But i had to use assertAlmostEqual when comparing the binCenters arrays because some error gets introduced... Is this ok?

- There's a use case that wouldn't be very used friendly: if histogram is used with default configuration (so bins = 10, minValue = 0 and maxValue = 1), 10 bins would be created between 0 and 1, and all values not between 1 and 0 would be disregarded. I feel reluctant overriding the minValue / maxValue settings of the user because I can't find out if they were explicitly set or just kept from the defaults..

Let me know..

test/src/unittests/stats/histogram/ranges.txt
Alia-morsi: the path is test/src/unittests/stats/stats/input.txt

src/algorithms/stats/histogram.cpp
dbogdanov: This sounds reasonable, let's go with 2)

test/src/unittests/stats/test_histogram.py
dbogdanov: I was referring to the parameters (min/max frequency, number of bins, etc.), not the input and output.

dbogdanov:
> Other Notes:
> 
>     * modified zero test to now include the 10 bin default. But i had to use assertAlmostEqual when comparing the binCenters arrays because some error gets introduced... Is this ok?
> 
>     * There's a use case that wouldn't be very user friendly: if histogram is used with default configuration (so bins = 10, minValue = 0 and maxValue = 1), 10 bins would be created between 0 and 1, and all values not between 1 and 0 would be disregarded. I feel reluctant overriding the minValue / maxValue settings of the user because I can't find out if they were explicitly set or just kept from the defaults..
> 
> 
> Let me know..

1) Using ``assertAlmostEqual`` should be fine
2) Just state in the documentation that the values outside the range are ignored.

test/src/unittests/stats/test_histogram.py
```diff
@@ -0,0 +1,62 @@
...
+# You should have received a copy of the Affero GNU General Public License
+# version 3 along with this program. If not, see http://www.gnu.org/licenses/
+
+
+from essentia_test import *
+
+class TestHistogram(TestCase):
+
+  def testZero(self):
+    histogram, binEdges = Histogram(normalize = "none", maxValue = 1., minValue = 0., numberBins = 10)(zeros(1000))
```
dbogdanov: No spaces around '=' in keyword arguments

dbogdanov:
Any updates on this?

Alia-morsi:
No unfortunately. I can attend to it starting the 15th of december, if
that's fine.

PR #801, opened 2018-11-23
Rtaudio support

pabloEntropia:
An algorithm to stream audio from the sound device using [RtAudio](https://github.com/thestk/rtaudio).


dbogdanov:
Merged into https://github.com/MTG/essentia/tree/rtaudio

TODOs:
- segfault on essentia.streaming import
- a python example

PR #807, opened 2018-12-19
Fix build system issues

Martchus:
I made these changes to build the library on the Raspberry Pi. The build system picks my CXX_FLAGS nicely. However, it also adds some unwanted ones.

Instead of adding more "clever" detections to decide when to use which flags this change just adds the possibility to skip unwanted ones. That's simple and sufficient.


Martchus:
I accidentally pushed the fix for libswresample (see https://github.com/MTG/essentia/issues/808) on the same branch. I can create a separate PR if required.

dbogdanov:
That's right, two separate pull requests will be better. 

Btw, #258 addresses builds on ARM, but having an explicit "no-msse" flag is a good option too.

Martchus:
Created a 2nd PR.

> Btw, #258 addresses builds on ARM, but having an explicit "no-msse" flag is a good option too.

It seems so. But it should work the other way around and only add those flags for architectures known to support them. I suppose the check for `arm` wouldn't even cover `aarch64` (because you would likely need to check for `arm64` there). So I decided to keep it simple rather than make it "clever" and just let the packager/user decide.

PR #822, opened 2019-01-30
Prevent the accidental (b or #) from being capitalized in ChordsDescriptors

pabloEntropia:
In the previous note naming convention sharp ("#") was used to name every not natural note and it was safe to capitalize the whole key string as "#" is not affected by `toUpper`.
Now the `Key` algorithm also outputs some flats ("b") so `toUpper` cannot be applied to the accidental anymore as "B" is not accepted for flat.

### DIFF
--- a/src/algorithms/tonal/chordsdescriptors.cpp
+++ b/src/algorithms/tonal/chordsdescriptors.cpp
@@ -102,7 +102,8 @@ void ChordsDescriptors::compute() {
     throw EssentiaException("ChordsDescriptors: Chords input empty");
   }
 
-  string key = toUpper(_key.get());
+  string key = _key.get();
+  key[0] = toUpper(string(1, key[0]))[0];
   string scale = toLower(_scale.get());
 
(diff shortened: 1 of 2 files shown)


alastair:
It'd be good to see a test for this too

src/algorithms/tonal/chordsdescriptors.cpp
```diff
@@ -102,7 +102,8 @@ void ChordsDescriptors::compute() {
     throw EssentiaException("ChordsDescriptors: Chords input empty");
   }
 
-  string key = toUpper(_key.get());
+  string key = _key.get();
+  key[0] = toUpper(string(1, key[0]))[0];
```
dbogdanov: Would that fail on an empty key input? I guess we might have a unit test for that.

pabloEntropia: Only the `chords` input was being checked. I added empty input checks for `key` and `scale` for consistency and unit tests for the 3 cases.

Issue #448, opened 2016-07-27
Make Accelerate default on OS X


carthach:
It's fast, stable and results in less dependencies. So we should prob set it as the default FFT on OS X in my view?

dbogdanov:
Sure!

dbogdanov:
This has been discussed in #826. For now, we would like to use FFTW for consistency whenever possible. Feel free to re-open in the future.

PR #826, opened 2019-02-05
Accelerate fixes

carthach:
Fixes Duration conflicts with Accelerate FFT library on Mac, also makes Accelerate default FFT library when compiling on Mac (meant to do that a long time ago).


dbogdanov:
Can you explain in more details the conflict with Duration?

carthach:
Compiling against the Accelerate framework includes MacTypes.h

https://opensource.apple.com/source/CarbonHeaders/CarbonHeaders-18.1/MacTypes.h.auto.html

Which already defines a typedef for Duration. Essentia code uses the 'using' keyboard quite a bit unfortunately, so it clashes. The easiest fix here is just to ensure the *FFTA headers are at the top of the algorithm_reg.

Hope you're well Dimi!

dbogdanov:
I can see sense in this proposed change, but I am not sure changing the default is a good idea for consistency of builds across platforms (e.g. building feature extractors, such as Music Extractor, or Python wheels, we may want to ensure specific dependencies to be used consistently across platform, this includes FFTW).  I would rather expect if one wants to use Accelerate to specify that explicitly. Please, provide more motivation :)

It will be nice to separate d9485b560964a5a039885f95d66108061b938da3 into a separate pull request.

carthach:
For the reasons stated here! https://github.com/MTG/essentia/issues/448

I accept your point about wanting a consistent standard across platforms, that's fair enough. I will separate the Accelerate fix into another pull request sure.

carthach:
The Duration issue seems to have resolved itself oddly... I'm closing.

dbogdanov:
Closing as we've cherry-picked the fix for the conflict with Duration in 51e8782949b04b617c5a763eb66076ed89a02193.

PR #828, opened 2019-02-06
NSGConstantQ: Update the default parameters and repeat the last sample in odd signals

pabloEntropia:
This PR contains a few small changes to improve the algorithm's performance with the default parameters.
- updated default parameters so they work better in most of the cases.
- repeat the last sample if the input signal is odd so FFT does not throw an exception.
- add a unit test to assert it don't throw exceptions on odd signals as before.

### DIFF
--- a/src/algorithms/standard/nsgconstantq.cpp
+++ b/src/algorithms/standard/nsgconstantq.cpp
@@ -279,11 +279,11 @@ void NSGConstantQ::compute() {
   }
 
   // FFT requires an even input, but in order to push back
-  // we needed to copy the data first. When it is alredy
+  // we needed to copy the data first. When it is already
   // even we can just use a reference to the input buffer
   // to avoid the copy.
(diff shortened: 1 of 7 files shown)


src/algorithms/standard/nsgconstantq.cpp
```diff
@@ -258,18 +258,34 @@ void NSGConstantQ::normalize() {
...
+  if (originalSignal.size() <= 1) {
     throw EssentiaException("NSGConstantQ: the size of the input signal is not greater than one");
   }
 
+  // FFT requires an even input, but in order to push back
+  // we needed to copy the data first. When it is alredy
+  // even we can just use a reference to the input buffer
+  // to avoid the copy.
+  if(originalSignal.size() % 2) {
+    E_INFO("Odd input. Repeating the last Sample.");
```
dbogdanov: "NSGConstantQ: Odd input size. Duplicating the last sample to get even size"

PR #830, opened 2019-02-11
Added one-liner function for HPCP in essentia.pytools

albincorreya:
Added a one-liner python function in essentia.pytools to compute [essentia.standard HPCP](http://essentia.upf.edu/documentation/reference/std_HPCP.html) features from an input audio vector.

```python
from essentia.standard import MonoLoader
import essentia.pytools.spectral as sp

audio_vector = MonoLoader("<path_to_your_audio_file>")
hpcp = sp.hpcpgram(audio_vector, frameSize=4096, hopSize=2048) # with default params
```

### DIFF
--- a/src/python/essentia/pytools/spectral.py
+++ b/src/python/essentia/pytools/spectral.py
@@ -246,72 +246,78 @@ def nsgcq_overlap_add(cq):
     return cqOverlap[:, hopSize:]
 
 
-def hpcp(audio, sampleRate=44100, frameSize=4096, hopSize=2048, windowType='blackmanharris62', 
-        harmonicsPerPeak=8, magnitudeThreshold=1e-05, maxPeaks=1000, whitening=True, numBins=12, 
-        referenceFrequency=440, minFrequency=40,maxFrequency=5000, nonLinear=False):
+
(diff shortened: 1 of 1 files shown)


dbogdanov:
Let's split hpcp function into two: 
- hpcpgram - computes hpcps on the entire audio input
- hpcp - computes hpcp given a spectrum frame

Use ```kwargs``` for additional keyword arguments and refer those to Essentia algorithms (HPCP, etc.) when possible.

src/python/essentia/pytools/spectral.py
```diff
@@ -243,3 +244,87 @@ def nsgcq_overlap_add(cq):
         cqOverlap[:, jj * hopSize + index] += cq[jj]
 
     return cqOverlap[:, hopSize:]
+
+
+def hpcp(audio, sampleRate=44100, frameSize=4096, hopSize=2048, windowType='blackmanharris62',
```
dbogdanov: Check if we can use ```**kwargs```` and refer to the algorithms documentation to declutter the list of all available parameters.

albincorreya: added a ```hpcpgram``` function which computes hpcp for overlapped frames of a given input audio signal.  Used ```**kwargs``` to pass additional parameters of essentia standard mode HPCP algorithm.

PR #845, opened 2019-03-25
3rdparty/cephes/bessel/bessel.h - include <cstddef> to avoid NULL not declared error

pcampr:
```
[284/287] Compiling src/3rdparty/cephes/bessel/iv.cpp
../src/3rdparty/cephes/bessel/iv.cpp: In function ‘double cephes_iv(double, double)’:
../src/3rdparty/cephes/bessel/iv.cpp:130:38: error: ‘NULL’ was not declared in this scope
  ikv_asymptotic_uniform(v, ax, &res, NULL);
                                      ^~~~
```
This error happens on Debian 9.6 with gcc 6.3. This doesn't happen e.g. on Ubuntu 18.04 nor 18.10.

The proposed fix adds `#include <cstddef>`, that declares `NULL`, to `bessel.h`, which is included in `iv.cpp` that generates the error.

### DIFF
--- a/src/3rdparty/cephes/bessel/bessel.h
+++ b/src/3rdparty/cephes/bessel/bessel.h
@@ -5,6 +5,7 @@
 //! \cond
 
 #include <cmath>
+#include <cstddef>
 
 #define MAXITER        500
 


dbogdanov:
We've just found and fixed that issue before seeing your message: 6d6c78a10ca72578e9250c5d58d1a7d7a2521300

PR #860, opened 2019-05-22
Make Degara beattracker deterministic

jpauwels:
Make Degara beattracker deterministic by not adding stochastic noise to silent frames

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,6 @@ packaging/win32_3rdparty/bin
 packaging/win32_3rdparty/share
 packaging/win32_3rdparty/lib
 packaging/win32_3rdparty/include
+dist
+essentia.egg-info
+tmp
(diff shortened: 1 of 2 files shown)


jpauwels:
Other candidate algorithms to check for same behaviour:

- TempoTapDegara
- BeatTrackerMultifeature
- RhythmExtractor

dbogdanov:
Thanks for this fix! 

Noise can be added to silent frames only in the streaming version of FrameCutter. For that reason, there's no problem for TempoTapDegara. We'll have to check the other ones.

dbogdanov:
Indeed BeatTrackerMultifeature also utilizes noise frames for silence, but I can't reproduce the problem.

.gitignore
```diff
@@ -84,3 +84,9 @@ packaging/win32_3rdparty/bin
 packaging/win32_3rdparty/share
 packaging/win32_3rdparty/lib
 packaging/win32_3rdparty/include
+test/src/unittest/*.npy
```
dbogdanov: This should be updated because all tests were moved to the `unittests`folder.

dbogdanov: In any case, we should rather not generate junk files inside `test/src/unit` test, but use a proper temp folder instead. I've created an issue for that (#861), so we can remove tose ignore lines here.

PR #890, opened 2019-07-25
ConstantQ refactor

dbogdanov:
Re-written erroneous implementation for the FFT-based ConstantQ algorithm and related SpectrumCQ and Chromagram algorithms.

Not tested yet, only merge after writing proper unit tests.


src/algorithms/standard/constantq.cpp
```diff
@@ -42,23 +42,25 @@ const char* ConstantQ::description = DOC("This algorithm computes Constant Q Tra
...
 void ConstantQ::compute() {
 
-  const vector<complex<Real> >& fft = _fft.get();
+  const vector<Real> & frame = _frame.get();
   vector<complex<Real> >& constantQ = _constantQ.get();
 
-  if (fft.size() != _inputFFTSize) {
-    throw EssentiaException("ConstantQ: input FFT size must be equal to: ", _inputFFTSize);
+  if (frame.size() != _windowSize) {
+    throw EssentiaException("ConstantQ: input FFT size must be equal to: ", _windowSize);
```
dbogdanov: --> "input frame size..."

src/algorithms/standard/constantq.cpp
```diff
@@ -70,76 +72,87 @@ void ConstantQ::configure() {
...
+  _fft->output("fft").set(_fftData);
 
   // Work only with a non-negative part of FFT as an input
-  _inputFFTSize = _windowSize/2+1;
+  _inputFFTSize = _windowSize / 2 + 1;
+
+  _sparseKernel = {};
 
-  struct SparseKernel &sk = _sparseKernel;
-  sk.i.reserve(_windowSize);
```
dbogdanov: What's the rationalle  not using `reserve()`? They are here to make interative ``push_back()`` calls faster.

dbogdanov:
```Chromagram``` algorithm should be adapted accordingly. It is missing the ``window`` parameter.

src/algorithms/standard/constantq.cpp
pabloEntropia: I just though that the code will be more clear without these references and forgot to keep the `reserves` when removing them. I'll put them back.

PR #901, opened 2019-08-27
MinMax algorithm to find the min or max value of an array

alastair:


src/algorithms/standard/minmax.cpp
```diff
@@ -0,0 +1,67 @@
...
+void MinMax::compute() {
+
+  const std::vector<Real>& input = _input.get();
+
+  switch (_type) {
+
+  case MIN:
+    {
+      if (input.empty())
+        throw EssentiaException("trying to get min of empty array");
```
alastair: the c++ implementation of mean does this check inside that function, I don't know if we should also move this check there?

src/algorithms/standard/minmax.cpp
```diff
@@ -0,0 +1,67 @@
...
+      _index.get() = argmin(input);
```
alastair: I'm not sure if it's best to use `.get()` here, or to get the index/input references earlier in the function like we do with _input

src/algorithms/standard/minmax.cpp
```diff
@@ -0,0 +1,67 @@
...
+      _value.get() = input[_index.get()];
+      break;
+    }
+
+  case MAX:
+    {
+      if (input.empty())
```
alastair: this code is almost duplicated, it could be moved into a macro like is done in UnaryOperator: https://github.com/MTG/essentia/blob/2456234775cb85ee5af91a80ed9b23a5618d2ab1/src/algorithms/standard/unaryoperator.cpp#L57

src/algorithms/standard/minmax.h
```diff
@@ -0,0 +1,95 @@
...
+
+#ifndef ESSENTIA_MINMAX_H
+#define ESSENTIA_MINMAX_H
+
+#include "algorithm.h"
+
+namespace essentia {
+namespace standard {
+
+class MinMax : public Algorithm {
```
alastair: Does it make sense to have a single algorithm with both behaviours, or do we prefer 2 separate algorithms?

test/src/unittests/stats/test_minmax.py
```diff
@@ -0,0 +1,71 @@
...
+from essentia.standard import MinMax
+
+from essentia_test import TestCase, TextTestRunner, allTests
+
+class TestMinMax(TestCase):
+
+    def testDefaultIsMin(self):
+        """Making the Algorithm with no type defaults to min"""
+        value, index = MinMax()([3,1,2])
+        self.assertEqual(value, 1.0)
```
alastair: value is always floating point, even if the inputs are only integers. Is this OK?

src/algorithms/standard/minmax.cpp
pabloEntropia: I agree that it would be good to move this check inside the `argmin` function so it would also cover other algorithms using the method.

src/algorithms/standard/minmax.cpp
pabloEntropia: Normally we take references of inputs and outputs at the beginning of `compute`. For example, 
https://github.com/MTG/essentia/blob/53c2fdb9544e8a8b6c5edcc22612a3ebc8eb619e/src/algorithms/spectral/mfcc.cpp#L112

I think it would improve readability a bit in this case.

src/algorithms/standard/minmax.cpp
pabloEntropia: This is a cool idea, but I don't think it is really required for a switch with 2 cases.

src/algorithms/standard/minmax.h
dbogdanov: I think it's fine to have both here.

src/algorithms/standard/minmax.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "MinMax.h"
+#include "essentiamath.h"
+#include <sstream>
+
+using namespace essentia;
+using namespace standard;
+
+const char* MinMax::name = "MinMax";
+const char* MinMax::category = "Standard";
+const char* MinMax::description = DOC("This algorithm calculates the minimum or maximum value of an array.");
```
dbogdanov: There may be more than one minimum or maximum value in the array. Please, specify the behavior in such a case in the DOC. For example, we use `std::max_element` and it specified that "If several elements in the range are equivalent to the greatest element, returns the iterator to the first such element."

test/src/unittests/stats/test_minmax.py
dbogdanov: The input to the algorithm is `std::vector<Real>` so those values get converted and we can't expect much more.

src/algorithms/standard/minmax.cpp
dbogdanov: We still may want to know in which algorithm the error has happen, so we can check here for EssentiaException and then provide a more informative message ("MinMax: ...")

test/src/unittests/stats/test_minmax.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def testMixedTypes(self):
+        value, index = MinMax()([4, 5, 3.3])
+        self.assertAlmostEqual(value, 3.3)
+        self.assertEqual(index, 2)
+
+    def testNonNumeric(self):
+        self.assertComputeFails(MinMax()(['a', 1]))
+
+    def testMax(self):
```
dbogdanov: Add a test for a correct index in the case of multiple min/max values in the array?

src/algorithms/standard/minmax.cpp
```diff
@@ -0,0 +1,67 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the Affero GNU General Public License
+ * version 3 along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+
+#include "MinMax.h"
```
dbogdanov: This results in an error in our Travis builds:
```
../src/algorithms/standard/minmax.cpp:20:20: fatal error: MinMax.h: No such file or directory
```

src/algorithms/standard/minmax.cpp
alastair: It seems like in other places in essentia once you catch an exception you output it to stderr. In this case I don't think this would work, because the behaviour of min will be undefined on an empty array. So should we catch the exception and re-raise it with the algorithm name prepended?

test/src/unittests/stats/test_minmax.py
alastair: we have testAllSame:
```py
value, index = MinMax()([0]*10)
```
but this doesn't include when there are multiple values spread around some other values. Do you want to add that one too?

alastair: added explicit tests for this case

src/algorithms/standard/minmax.cpp
alastair: yes, we did this

PR #911, opened 2019-09-13
Update `print` and `except` calls for python3 support

mmercedes:


alastair:
Looks good, we should also add `from __future__ import print_function` so that it continues to work in python 2.

mmercedes:
good call @alastair

Issue #806, opened 2018-12-14
Update instructions on building with emscripten


mmxgn:
Hello,

First of all, essentia is a great project and I have used in extensively on python. Thanks everyone involved.

I was trying to build essentia with emscripten but I have run in a variety of problems that I am unable to pinpoint and solve on my own.

I have tested on Ubuntu 18.04 (LTS), Archlinux and the `trzeci/emscripten` docker container.

With essentia-2.1_beta3 (the only one I managed to get it to compile to some point) I get:
```
 emconfigure sh -c './waf configure --prefix=$EMSCRIPTEN/system/local/ --lightweight=fftw --emscripten'

emmake ./waf

...
../src/algorithms/standard/constantq.cpp:65:29: error: implicit conversion from '_Complex double' to 'float' is not permitted in C++
  constantQ.assign(_uK, 0.0 + 0.0j); // initialize output
                        ~~~~^~~~~~
../src/algorithms/standard/constantq.cpp:104:58: error: implicit conversion from '_Complex double' to 'double' is not permitted in C++
  vector<complex<double> > hammingWindow(_FFTLength, 0.0 + 0.0j);
                                                     ~~~~^~~~~~
../src/algorithms/standard/constantq.cpp:105:63: error: implicit conversion from '_Complex double' to 'float' is not permitted in C++
  vector<complex<Real> > transfHammingWindowR(_FFTLength, 0.0 + 0.0j);
                                                          ~~~~^~~~~~
../src/algorithms/standard/constantq.cpp:120:42: error: implicit conversion from '_Complex double' to 'double' is not permitted in C++
    hammingWindow.assign(_FFTLength, 0.0 + 0.0j);
                                     ~~~~^~~~~~
4 errors generated.
ERROR    root: compiler frontend failed to generate LLVM bitcode, halting
```

Trying using the master branch I get:

```
...
../src/essentia/streaming/algorithms/../../pool.h:574:1: note: in instantiation of member function 'std::__1::vector<std::__1::vector<float, std::__1::allocator<float> >, std::__1::allocator<std::__1::vector<float, std::__1::allocator<float> > > >::operator=' requested here
SPECIALIZE_APPEND(std::vector<Real>, VectorReal);
^
../src/essentia/streaming/algorithms/../../pool.h:569:22: note: expanded from macro 'SPECIALIZE_APPEND'
  _pool##tname[name] = values;                                                        \
                     ^
note: candidate function not viable: requires 1 argument, but 2 were provided
[...]

PR #913, opened 2019-09-26
update waf to 2.0.18 with emscripten compiler support 

albincorreya:
As detailed in issue #806, the existing ./waf file fails to detect the emscripten compiler. 
This PR adds 
- a new waf file with emscripten compiler support which is regenerated from the [main waf repo](https://gitlab.com/ita1024/waf) using the command `./waf-light --tools=c_emscripten`. 

- also updates the waf version to 2.0.18


alastair:
In gaia when we recompiled with additional tools, we added instructions to the commit message as a way of adding documentation about how to re-run it in the future:
https://github.com/MTG/gaia/commit/8b884cc07e337f2ffba078c9ec5f1dbd6c69cfcd
Maybe we could do the same here? Having it in github is good, but I'd like to also see it in the actual repo.

albincorreya:
@alastair, done!

PR #931, opened 2019-12-05
fix dead documentation links

gregsadetsky:


alastair:
Thanks for the contribution. This page is rendered into html at https://essentia.upf.edu/documentation/extractors_out_of_box.html, and therefore the links should point to the rendered .html version of the pages. Do you know if sphinx will rewrite these links when generating the html documentation?

dbogdanov:
> Do you know if sphinx will rewrite these links when generating the html documentation?

No, it won't. We need to find a correct way to do that.

PR #802, opened 2018-11-26
Tensorflow support

pabloEntropia:
Set of algorithms to support Tensorflow C API inside Essentia. 

The Algorithm TensorflowPredict accepts Tensorflow graphs in [Protobuff](https://www.tensorflow.org/serving/serving_basic) format.

Streaming prediction is also supported with the helper functions: VectorRealToTensor and TensorToPool. TensorflowPredict returns a stream of pools that can be converted to a stream of frames with the helper functions PoolToTensor and TensorToVectorReal.
A typical pipelines would be:
`AudioLoader >> FrameCutter >> Windowing >> Spectrum >> MelBands >> VectorRealToTensor >> TensorToPool >> TensorFlowPredict >> PoolToTensor >> TensorToVectorReal`
This architecture allows retrieving intermediate layers of the graph and saving  them in the output pools. Thus, different `PoolToTensor` instances could be used to extract the ouput of different layers if needed.

This PR introduces Boost::multi_array as the base type for tensors in Essentia.

PR #933, opened 2019-12-09
Tensorflow support with eigen3

pabloEntropia:
Reimplementation of #802 relying on `eigen 3` for the tensor management.


src/algorithms/standard/vectorrealtotensor.h
```diff
@@ -0,0 +1,73 @@
...
+  VectorRealToTensor(){
+    declareInput(_frame, 128,"frame", "the input frames");
+    declareOutput(_tensor, 1, "tensor", "the accumulated frame in one single tensor");
+
+  }
+
+  void declareParameters() {
+    // TODO: set a better default shape
+    std::vector<int> outputShape = {-1, 1, 128, 128};
+    declareParameter("shape", "the size of output tensor. If batch dimension (the first one) is -1 it will accumulate as many batches as available in the input track", "", outputShape);
```
dbogdanov: We can add more hints on what dimensions mean: "the size of output tensor (batch size, number of channels, ... )"

src/algorithms/standard/vectorrealtotensor.cpp
```diff
@@ -0,0 +1,227 @@
...
+using namespace std;
+
+namespace essentia {
+namespace streaming {
+
+const char* VectorRealToTensor::name = "VectorRealToTensor";
+const char* VectorRealToTensor::category = "Standard";
+const char* VectorRealToTensor::description = DOC("This algorithm takes a stream of frames "
+"and outputs them as a single tensor everytime the output shape is reached. ")
+"If the batch dimension (0) is set to -1 the data will be accumulated and returned "
```
dbogdanov: Explain terminology better: what is the difference between batch, patch, and frames.

src/algorithms/standard/vectorrealtotensor.h
dbogdanov: "If batch dimension (the first one) is -1 it will accumulate as many batches as available in the input track" -- rephrase

src/algorithms/standard/vectorrealtotensor.h
```diff
@@ -0,0 +1,73 @@
...
+    declareParameter("patchHopSize", "number of frames between the beginnings of adjacent patches. 0 to avoid overlap", "[0,inf)", 0);
+    declareParameter("batchHopSize", "number of patches between the beginnings of adjacent batches. 0 to avoid overlap", "[0,inf)", 0);
+    declareParameter("lastPatchMode", "what to do with the last incomplete patch", "{discard,repeat}", "repeat");
```
dbogdanov: repeat = repeat the last frame?

src/algorithms/standard/vectorrealtotensor.h
```diff
@@ -0,0 +1,73 @@
...
+
+  static const char* name;
+  static const char* category;
+  static const char* description;
+};
+
+} // namespace streaming
+} // namespace essentia
+
+#endif // ESSENTIA_VECTORREALTOTENSOR_H
```
dbogdanov: There's a few newline warnings like this for some files

src/algorithms/machinelearning/tensorflowpredict.h
```diff
@@ -0,0 +1,134 @@
...
+    declareParameter("graphFilename", "the name of the file from which to read the Tensorflow graph", "", "/home/pablo/base_model.pb");
+
+    const char* inputNames[] = {"input_1"};
+    const char* outputNames[] = {"output_node0"};
+
+    std::vector<std::string> inputNamesVector = arrayToVector<std::string>(inputNames);
+    std::vector<std::string> outputNamesVector = arrayToVector<std::string>(outputNames);
+
+    declareParameter("inputs", "will look for this namespaces in poolIn. Should match the names of the input nodes in the Tensorflow graph", "", inputNamesVector);
+    declareParameter("outputs", "will save the tensors on the graph nodes named after `outputs` to the same namespaces in the output pool", "", outputNamesVector);
```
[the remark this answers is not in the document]
alastair: something like this: ?
inputs -> a list of names of input nodes in the Tensorflow graph to read from poolIn
outputs -> a list of names output graph nodes from the Tensorflow graph to save to the output pool

I don't quite understand what these arguments do, come and talk to me if you want to explain in more detail

src/algorithms/standard/vectorrealtotensor.h
alastair: "the size of the output tensor (batchsize, arg2, arg3, arg4). If batchsize is -1, all batches from the input track will be (used/gathered)" ?

pabloEntropia:
After some tries I found a way to run the 3 Travis  jobs under the log length and time constrains and
4e57863 is passing now. Maybe this could be a temporal solution until we decide if we want to split the process into testing/wheel generation.

pabloEntropia:
Some Travis checks failed on deploy time with error,
`(56) GnuTLS recv error (-54): Error in the pull function.`
Running `curl` with `--retry` as suggested [here](https://stackoverflow.com/questions/42873285/curl-retry-mechanism) may fix it.

PR #938, opened 2020-01-09
Drop support for Python 3.4 wheels

pabloEntropia:
We are dropping support for Python 3.4 since PyYaml is not supporting it anymore.
We can just remove the Python3.4 folder until ManyLinux1 drops the support too. 
This way, we prevent making skip conditions in two 2 different loops so the code doesn't get too messy.

### DIFF
--- a/travis/build_wheels.sh
+++ b/travis/build_wheels.sh
@@ -14,6 +14,10 @@ set -e -x
 #PKG_CONFIG_PATH=/usr/local/lib/pkgconfig
 #PKG_CONFIG_PATH=/io/packaging/debian_3rdparty/lib/pkgconfig
 
+# We are dropping support for Python 3.4 since PyYaml is not supporting it anymore.
+# We can just remove the Python3.4 folder until ManyLinux1 drops the support too.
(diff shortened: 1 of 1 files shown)


alastair:
what does it mean for pyyaml to not support it? Does it mean that they are using a python feature in their latest version which is only available in python 3.5? or have they only said this and not yet used any new things?

pabloEntropia:
It means that they [explicitly ask](https://github.com/yaml/pyyaml/commit/a7a97871fce027d8c3da0d96117f5b71536b8244#diff-2eeaed663bd0d25b7e608891384b7298) for a Python 3 version > 3.4 in their setup script.

After some discussion, we concluded that it makes sense to drop support for Python 3.4 provided that:
1. Most of the python libraries around Essentia are dropping Python 3.4 for their latest versions (Numpy, PyYaml, Tensorflow...)
2. Ubuntu 1.14 LTS is the last one relying on Python 3.4 but it's already out of the maintenance updates phase.
3. It's easy to patch Essentia's setup script to depend on `pyyaml==5.2` in case of need for Python 3.4

PR #941, opened 2020-01-15
Cover similarity algorithms

dbogdanov:


test/src/unittests/highlevel/test_chromacrosssimilarity.py
```diff
@@ -43,49 +41,47 @@ class TestChromaCrossSimilarity(TestCase):
...
     def testRegressionStreaming(self):
         """Tests streaming ChromaCrossSimilarity algo against the standard mode algorithm with 'streamingMode=True' """
         # compute chromacrosssimilarity matrix using standard mode
-        csm_standard = estd.ChromaCrossSimilarity(streamingMode=True)
-        sim_matrix_std = csm_standard.compute(self.query_hpcp, self.reference_hpcp)
+        csm_standard = ChromaCrossSimilarity(streaming=True)
+        sim_matrix_std = csm_standard(self.query_hpcp, self.reference_hpcp)
         # compute chromacrosssimilarity matrix using streaming mode
         queryVec = ess.VectorInput(self.query_hpcp)
         csm_streaming = ess.ChromaCrossSimilarity(referenceFeature=self.reference_hpcp, oti=0)
```
dbogdanov: --> `oti=False` for type consistency in the code

test/src/unittests/highlevel/test_coversongsimilarity.py
```diff
@@ -42,27 +38,15 @@ def testEmpty(self):
     def testRegressionStandard(self):
         sim = CoverSongSimilarity()
         score_matrix, distance = sim.compute(array(self.sim_matrix))
-        self.assertAlmostEqual(self.expected_distance, distance)
+        self.assertAlmostEqual(self.expected_distance, np.round(distance, 3))
```
dbogdanov: A proper way to do it is with `assertAlmostEqualFixedPrecision`

test/src/unittests/highlevel/test_crosssimilaritymatrix.py
```diff
@@ -18,7 +18,6 @@
 # version 3 along with this program. If not, see http://www.gnu.org/licenses/
 
 from sklearn.metrics.pairwise import euclidean_distances
```
dbogdanov: This is the only place where sklearn is used in Essentia. We should at least replace it for [scipy](with https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance_matrix.html), or even re-implement from scratch. 

Even better, we can also use a pre-computed answer. In the case we use simple matrices in the multiplication that's easy to hardcode in the code. Otherwise, load from a file. Are current query and reference features in the tests the best to detect possible issues? One in the matrix of ones, another one is matrix of zeros.

test/src/unittests/highlevel/test_crosssimilaritymatrix.py
```diff
@@ -51,22 +50,21 @@ class TestCrossSimilarityMatrix(TestCase):
...
-        self.assertAlmostEqualVector(self.expected, result)
+        self.assertAlmostEqualMatrix(self.expected, result)
 
     def testRegressionBinary(self):
         csm = CrossSimilarityMatrix(binarize=True, binarizePercentile=0.095)
         result = csm.compute(self.query_feature, self.reference_feature)
-        self.assertAlmostEqualVector(self.expected_binary, result)
+        self.assertAlmostEqualMatrix(self.expected_binary, result)
 
     def testInvalidParam(self):
```
dbogdanov: Remove this test, if it is only a "pass".

test/src/unittests/highlevel/test_crosssimilaritymatrix.py
albincorreya: Replaced query and reference hpcp matrix values with the ones in `test_chromacrosssimilaritymatrix. Added pre-computed result for the cross similarity matrix and thus removed the extra overhead of computing it in the test as you suggested (also the sklearn dependency).

PR #943, opened 2020-01-15
MusiCNN and VGGish algorithms

pabloEntropia:
This PR contains:
-  dedicated algorithms for input feature extraction and prediction with models based on the MusiCNN and VGGish architectures.
- C++ extractors for those.
- Regression unit tests to compare with the original Python implementation.
- A submodule [essentia-models](https://github.com/MTG/essentia-models) to store models for testing purposes.


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "test/audio"]
 	path = test/audio
 	url = https://github.com/MTG/essentia-audio
-[submodule "essentia-models"]
+[submodule "essentia/models"]
```
dbogdanov: If we want to store the models inside `test` folder, the name of submodule should better be `test/models`. Also missing a  `git add` for this `test/models` folder.

src/algorithms/machinelearning/tensorflowpredict.h
```diff
@@ -85,9 +85,9 @@ class TensorflowPredict : public Algorithm {
 
     declareParameter("inputs", "will look for this namespaces in poolIn. Should match the names of the input nodes in the Tensorflow graph", "", inputNamesVector);
     declareParameter("outputs", "will save the tensors on the graph nodes named after `outputs` to the same namespaces in the output pool", "", outputNamesVector);
-    declareParameter("isTraining", "Whether to return the output in training mode (normalized with statistics of the current batch) or in inference mode (normalized with moving statistics)", "{true,false}", false);
-    declareParameter("isTrainingName", "The name if the input reciving isTraining. If empty, it wont be feed ro the network", "", "");
-    declareParameter("squeeze", "Remove singleton dimensions of the inputs tensors. Does not apply to the batch dimension.", "{true,false}", true);
+    declareParameter("isTraining", "run the model in training mode (normalized with statistics of the current batch) or in inference mode (normalized with moving statistics). It only applies to some models", "{true,false}", false);
```
dbogdanov: "It only applies" --> "This only applies"

dbogdanov: " or in inference mode" --> "instead of inference mode" ? Because the parameter value is true or false.

src/algorithms/machinelearning/tensorflowpredictmusicnn.h
```diff
@@ -56,7 +56,7 @@ class TensorflowPredictMusiCNN : public AlgorithmComposite {
     declareParameter("graphFilename", "the name of the file containing the model to use", "", Parameter::STRING);
     declareParameter("input", "the name of the input nodes in the Tensorflow graph", "", "model/Placeholder");
     declareParameter("output", "the name of the node from which to retrieve the output tensors", "", "model/Sigmoid");
-    declareParameter("isTrainingName", "The name if the input reciving isTraining. If empty, it wont be feed ro the network", "", "");
+    declareParameter("isTrainingName", "the name of an additional input node to indicate the model if it is in training mode or not. Leave it empty when the model does not need such input", "", "");
```
dbogdanov: Still not sure, perhaps this will be an improvement? 

"the name of an additional input node indicating whether the model is to be run in a training mode (for models with a training mode, leave it empty otherwise)"

src/algorithms/machinelearning/tensorflowpredictvggish.h
```diff
@@ -56,7 +56,7 @@ class TensorflowPredictVGGish : public AlgorithmComposite {
     declareParameter("graphFilename", "the name of the file containing the model to use", "", Parameter::STRING);
     declareParameter("input", "the name of the input nodes in the Tensorflow graph", "", "model/Placeholder");
     declareParameter("output", "the name of the node from which to retrieve the output tensors", "", "model/Sigmoid");
-    declareParameter("isTrainingName", "The name if the input reciving isTraining. If empty, it wont be feed ro the network", "", "");
+    declareParameter("isTrainingName", "the name of an additional input node to indicate the model if it is in training mode or not. Leave it empty when the model does not need such input", "", "");
```
dbogdanov: Same here

PR #953, opened 2020-02-04
Tensorflowpredict improvements

pabloEntropia:
Some improvements for TensorflowPredict.
- Initialize TensorFlow variables in the constructor
- Call TF_DeleteSession from reset to free the resources of the previous session.
- Modify configure to delete the previous session and graph before making new ones.
- Print all the nodes available in the graph when the input/output node
  name is not found to help to find the correct one.


pabloEntropia:
I tested the proposed changes using TensorFlowpredictVGGish with [danceability-vggish-audioset ](https://essentia.upf.edu/models/classifiers/danceability/danceability-vggish-audioset.pb) in a compute-reset loop of 100 iterations running on CPU and GPU. In both cases, the process shows a stable memory footprint after the changes. Before, the reset method was creating a new session without freeing the previous one generating a memory leakage.

src/algorithms/machinelearning/tensorflowpredict.cpp
```diff
@@ -293,8 +311,28 @@ TF_Output TensorflowPredict::graphOperationByName(const char* nodeName,
   TF_Output output = {TF_GraphOperationByName(_graph, nodeName), index};
 
   if (output.oper == nullptr) {
-    throw EssentiaException("TensorflowPredict: Can't init node names.");
+    vector<string> nodes = TensorflowPredict::nodeNames();
+
+    string exceptionText = "TensorflowPredict: '" + string(nodeName) + "' is not a valid node name of this graph.\n" +
+                           "Available node names are:\n";
+
+    for (vector<string>::const_iterator i = nodes.begin(); i != nodes.end() - 1; ++i) exceptionText +=  *i + ", ";
```
dbogdanov: Looks like this code doubles E_INFO output by the `configure()` method

src/algorithms/machinelearning/tensorflowpredict.h
```diff
@@ -76,15 +84,8 @@ class TensorflowPredict : public Algorithm {
...
-
-    const char* inputNames[] = {"input_1"};
-    const char* outputNames[] = {"output_node0"};
-
-    std::vector<std::string> inputNamesVector = arrayToVector<std::string>(inputNames);
-    std::vector<std::string> outputNamesVector = arrayToVector<std::string>(outputNames);
-
-    declareParameter("inputs", "will look for this namespaces in poolIn. Should match the names of the input nodes in the Tensorflow graph", "", inputNamesVector);
-    declareParameter("outputs", "will save the tensors on the graph nodes named after `outputs` to the same namespaces in the output pool", "", outputNamesVector);
+    declareParameter("inputs", "will look for this namespaces in poolIn. Should match the names of the input nodes in the Tensorflow graph", "", Parameter::VECTOR_STRING);
```
dbogdanov: type: "this namespaces" --> "these namespacese"

src/algorithms/machinelearning/tensorflowpredict.cpp
pabloEntropia: When the first element of `inputs` is an empty string the list of available nodes is printed through E_INFO, but when it is an invalid node name (not existing in the graph) this list is printed as part of the exception message.
It's true that there are a couple of duplicated lines, but this permits to have the method `nodeNames` returning a vector<string> which is more flexible than returning the list of nodes as a single string and it may be useful in the future.

PR #814, opened 2019-01-17
Key updates

dbogdanov:
- Update Key (add detuning correction, low-energy HPCP bin thresholding, new key profiles 'braw' and 'bgate' instead of 'faraldo' and 'pentatonic')
- Update KeyExtractor (change default parameters, add spectral whitening and tuning correction steps)
- New algorithm NNLSChroma (derived from the original NNLS Chroma code)
- New algorithm LogSpectrum (derived from the original NNLS Chroma code)
- Documentation updates

This makes obsolete previous PR #614

PR #955, opened 2020-02-06
Freesound Extractor update

ffont:


src/algorithms/tonal/key.cpp
```diff
@@ -77,7 +77,9 @@ void Key::configure() {
       _useMajMin = false;
     }
   }
-  const char* keyNames[] = { "A", "Bb", "B", "C", "C#", "D", "Eb", "E", "F", "F#", "G", "Ab" };
+  // TODO: choose which set of key names we want to use
+  //const char* keyNames[] = { "A", "Bb", "B", "C", "C#", "D", "Eb", "E", "F", "F#", "G", "Ab" };
+  const char* keyNames[] = { "A", "A#", "B", "C", "C#", "D", "D#", "E", "F", "F#", "G", "G#" };
```
dbogdanov, line 82: We should be careful about this change as it would break compatibility with previous versions of Essentia. We previously used the key names proposed in this change but we switched from them in https://github.com/MTG/essentia/pull/814. Let's discuss this again. @palonso @angelfaraldo 

Related: https://github.com/MTG/essentia/pull/815

Maybe this can be resolved by having a parameter that defines a key name style to use.

src/essentia/streaming/streamingalgorithm.cpp
```diff
@@ -33,7 +33,9 @@ void connect(SourceBase& source, SinkBase& sink) {
     //     this is a requirement for ProxyConnectors
     E_DEBUG(EConnectors, "Connecting " << source.fullName() << " to " << sink.fullName());
     sink.connect(source);
+    E_DEBUG(EConnectors, "  " << source.fullName() << " ok");
     source.connect(sink);
+    E_DEBUG(EConnectors, "  " << sink.fullName() << " ok");
```
dbogdanov, line 38: Are these new `E_DEBUG`s necessary? They clutter the output.

test/src/unittests/essentia_test.py
```diff
@@ -183,6 +181,40 @@ def assertAlmostEqualAudio(self, found, expected, precision = 1e-7):
...
           self.assertAlmostEqualAudio(val1, val2, precision)
 
+    def assertAlmostEqualPool(self, pool_found, pool_expected, precision = 1e-7):
+        # Assert both pools have the same descriptors
+        self.assertListEqual(sorted(pool_found.descriptorNames()), sorted(pool_expected.descriptorNames()))
+
+        # Compare descriptor values one by one
+        for descriptorName in pool_found.descriptorNames():
+            if 'essentia_git_sha' in descriptorName:
+                # Do not compare descriptor 'essentia_git_sha' as it will aleays be different when essentia is built
```
dbogdanov, line 191: misspell: "always"

dbogdanov (changes requested):

doc/sphinxdoc/freesound_extractor.rst
```diff
@@ -107,7 +107,9 @@ For implementation details, see `the code of extractor <https://github.com/MTG/e
 
 * **beats_count**: number of detected beats
 
-* **bpm**: beats per minute (BPM) value according to detected beats
+* **bpm**, **bpm_confidence**: BPM value according to detected beats and confidence value for the prediction. Note that the confidence values goes from 0 to 5.32. Algorithms: Algorithms: `RhythmExtractor2013 <reference/streaming_RhythmExtractor2013.html>`_
+
+* **bpm_loop**, **bpm_loop_confidence**: BPM value according to detected beats using an algorithm specilized in loops, and confidence value for the prediction. Note that the confidence values goes from 0 to 1. A confidence of 1 could be used to determine whether the sound is a loop or not. Algorithms: Algorithms: `LoopBpmEstimator <reference/streaming_LoopBpmEstimator.html>`_, `LoopBpmConfidence <reference/streaming_LoopBpmConfidence.html>`_
```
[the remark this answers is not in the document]
ffont: Addressed in #da8a3b64c40262fbadd5bf5509e33d6c37916cdb, also found other duplicates and fixed

src/algorithms/tonal/key.cpp
ffont: Addressed in 3089d2d8332ad24eae5bdd59e4bae1b1de134971

src/essentia/utils/extractor_freesound/FreesoundTonalDescriptors.cpp
```diff
@@ -87,97 +89,35 @@ void FreesoundTonalDescriptors ::createNetwork(SourceBase& source, Pool& pool) {
...
+                                           "hopSize", frameSize,  // No overlapp (that's how KeyExtractor sets the defaults)
+                                           "hpcpSize", 12);
+  Algorithm* high_pass_1 = factory.create("HighPass",
+                                          "cutoffFrequency", 200.);
+  Algorithm* high_pass_2 = factory.create("HighPass",
+                                          "cutoffFrequency", 200.);
+  Algorithm* high_pass_3 = factory.create("HighPass",
+                                          "cutoffFrequency", 200.);
+  source >> high_pass_1->input("signal");
+  high_pass_1->output("signal") >> high_pass_2->input("signal");
```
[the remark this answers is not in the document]
ffont: Absolutely no idea, but this is only used to feed the key extractor. Maybe we should just remove it and have the same preprocessing for the key extractor as done in the music extractor.

ffont:
Thanks for the comments. I'll look at the unit tests and remove the weird pre-processing for key extractor.

src/algorithms/sfx/tctototal.cpp
```diff
@@ -28,8 +28,8 @@ const char* TCToTotal::name = "TCToTotal";
 const char* TCToTotal::category = "Envelope/SFX";
 const char* TCToTotal::description = DOC("This algorithm calculates the ratio of the temporal centroid to the total length of a signal envelope. This ratio shows how the sound is 'balanced'. Its value is close to 0 if most of the energy lies at the beginning of the sound (e.g. decrescendo or impulsive sounds), close to 0.5 if the sound is symetric (e.g. 'delta unvarying' sounds), and close to 1 if most of the energy lies at the end of the sound (e.g. crescendo sounds).\n"
 "\n"
-"Please note that the TCToTotal ratio is not defined for a zero signal (a signal consisting of only zeros), nor it is defined for a signal of less than 2 elements."
-"An exception is thrown if the given envelope's size is not larger than 1. And also if the integral of the input envelope is 0 (i.e. envelope is only zeros or if its sum is 0).\n\n"
+"Please note that the TCToTotal ratio will return 0.5 for a zero signal (a signal consisting of only zeros) as 0.5 is the middle point of the signal. TCToTotal is not defined for a signal of less than 2 elements."
```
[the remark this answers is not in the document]
ffont: I updated the tests so that they match this change in ecd3f5213e057194c56aa8aaa8f040e3401ff0c3
Also I fixed a bug in TCToTotal streaming that would not return 0.5 when signal is empty (and would return nan instead, an early return was missing).

src/essentia/streaming/streamingalgorithm.cpp
ffont: No idea what this is. Removed in 63634927f0b3cf1bfd14ae7184ffc770e940ff17

ffont:
Thanks for the review @dbogdanov. I made some more changes and replied to your comments :)

src/essentia/utils/extractor_freesound/FreesoundTonalDescriptors.cpp
```diff
@@ -87,97 +89,35 @@ void FreesoundTonalDescriptors ::createNetwork(SourceBase& source, Pool& pool) {
...
-  hpcp_chord->output("hpcp") >> schord->input("pcp");
-  schord->output("chords") >> PC(pool, nameSpace + "chords_progression");
-  schord->output("strength") >> PC(pool, nameSpace + "chords_strength");
-  
-  Algorithm* schords_desc = factory.create("ChordsDescriptors");
-  schord->output("chords") >> schords_desc->input("chords");
-  skey_temperley->output("key") >> schords_desc->input("key");
-  skey_temperley->output("scale") >> schords_desc->input("scale");
+  // Compute key
+  Algorithm* key_extractor = factory.create("KeyExtractor",
```
dbogdanov, line 98: @ffont This is using the default `profileType` key profile from KeyExtractor. We can be more explicit and manually configure it here, if you need to ensure using a specific key profile.

src/essentia/utils/extractor_freesound/FreesoundTonalDescriptors.cpp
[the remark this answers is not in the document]
dbogdanov: I guess this is still pending 🙂
Maybe you can leave a comment in the code if there is any reference/rationale for using the high pass?

dbogdanov (changes requested):
Looks good!
This is the final review with two minor comments.

src/essentia/utils/extractor_freesound/FreesoundTonalDescriptors.cpp
ffont: No need of a specific profile, let's use the default.

src/essentia/utils/extractor_freesound/FreesoundTonalDescriptors.cpp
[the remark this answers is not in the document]
ffont: If this is not done in the music extractor, let's remove it from here as well as I have no idea why this is here. Maybe it was a mistake Gerard made many years ago. I can't think of a reason why we'd want that unless this is also being done in the other extractors.

src/essentia/utils/extractor_freesound/FreesoundTonalDescriptors.cpp
dbogdanov: The profiles we previously used are: `temperley`, `krumhansl`, `edma`. KeyExtractor defaults to `bgate`.

src/essentia/utils/extractor_freesound/FreesoundTonalDescriptors.cpp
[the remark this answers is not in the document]
dbogdanov: Sound good, let's remove it then!

PR #960, opened 2020-02-19
Fix broken links (.html -> .rst)

philipkobernik:


alastair:
Thanks for the contribution. Please see #931, we're already trying to work out a good way to handle this in our documentation

PR #983, opened 2020-05-12
essentia-tensorflow wheels for Python 3.8

palonso:
This PR modifies `build_wheels.py` to build essentia-tensorflow wheels for Python 3.8 (the default one in Ubuntu 20.04) with TensorFlow 2.2.0.

Additionally, the script that generates the TensorFlow symbolic links is imported.


src/3rdparty/tensorflow/setup_tensorflow.py
```diff
@@ -74,18 +74,27 @@
...
 
-        src = join(tf_dir, 'libtensorflow_framework.so.1')
-        tgt = join(context, 'lib', 'libtensorflow_framework.so')
+        libtensorflow = 'tensorflow_framework'
+        pywrap_tensorflow_internal = 'pywrap_tensorflow_internal'
+
+        libtensorflow_name = 'lib{}.so.{}'.format(libtensorflow, major_version)
+        src = join(tf_dir, libtensorflow_name)
+        tgt = join(context, 'lib', libtensorflow_name)
         call(['ln', '-sf', src, tgt])
```
dbogdanov, line 83: We can use [`os.symlink`](https://docs.python.org/3/library/os.html#os.symlink) instead

pabloEntropia: I didn't use it because it lacks the `--force` flag and it raises an exception if the target exists. If we decide to use `os.symlink` we should check and remove the target if it exists manually before. Should we make a small function based on `os` methods for this?

dbogdanov: Something like this would work: https://mail.python.org/pipermail/python-list/2000-March/055125.html

PR #1002, opened 2020-06-23
Fix cross compilation for Android

amCap1712:
Currently, essentia builds invalid binaries while cross compiling for Android. This is because the generic clang compiler is used instead of the one available in Android NDK toolchain. Also, either an ABI specific clang compiler provided in the toolchain should be used or the -target flag must be passed to the android specific clang compiler. These changes have been made in accordance with the [official NDK documentation](https://developer.android.com/ndk/guides/other_build_systems#overview).

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -125,11 +125,11 @@ Build Essentia with static examples:
 Cross-compiling for Android
 ---------------------------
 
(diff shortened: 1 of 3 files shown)


amCap1712:
The build-android branch makes this oboslete.

Issue #914, opened 2019-10-02
failing Python tests (after fixing some problems)


boegel:
While running the Essentia Python tests with `python ./waf run_python_tests`, I ran into a couple of problems.

Context:

* GCC 8.2 (compilation flags: `-O2 -ftree-vectorize -march=native -fno-math-errno`)
* Essentia 2.1_beta5
* Python 2.7.15 + numpy 1.16.2
* FFTW 3.3.8
* FFmpeg 4.1.3
* libsamplerate 0.1.9
* TagLib 1.11.1
* Chromaprint 1.4.3
* Gaia 2.4.5

First, I had apply a small patch to ensure that `$PYTHONPATH` doesn't get wiped when running the Python tests (we install `numpy` in a separate location, and set `$PYTHONPATH` to tell Python where to find it):

```diff
--- essentia-2.1_beta5/wscript.orig	2019-10-01 20:26:06.136279484 +0200
+++ essentia-2.1_beta5/wscript	2019-10-01 20:26:57.286223056 +0200
@@ -338,7 +338,7 @@
     os.system('cp -r src/python/essentia build/python/')
     os.system('cp build/src/python/_essentia*.so build/python/essentia')

-    ret = os.system('PYTHONPATH=build/python %s test/src/unittests/all_tests.py' % sys.executable)
+    ret = os.system('PYTHONPATH=build/python:$PYTHONPATH %s test/src/unittests/all_tests.py' % sys.executable)
     if ret:
         ctx.fatal('failed to run python tests. Check test output')
```

Then, I had to rename `test/src/unittests/io/` to `test/src/unittests/iotest/` to avoid this problem with the `io.open` in `numpy` getting broken (numpy 1.16.2):

```
Unable to import NumPy C API from Essentia module. Error code = -1
Traceback (most recent call last):
  File "test/src/unittests/all_tests.py", line 27, in <module>
    import essentia
  File "/tmp/essentia-2.1_beta5/build/python/essentia/__init__.py", line 1, in <module>
    from . import _essentia
  File "/software/SciPy-bundle/2019.03-foss-2019a/lib/python2.7/site-packages/numpy/__init__.py", line 145, in <module>
    from . import lib
  File "/software/SciPy-bundle/2019.03-foss-2019a/lib/python2.7/site-packages/numpy/lib/__init__.py", line 24, in <module>
    from .npyio import *
[...]

PR #1003, opened 2020-06-25
Waf run_python_tests fixes

pabloEntropia:
A couple of fixes for the `run_python_test` Waf method:

- Replace 'is not' by '!=' for literals, which throws a warning in Python 3.8.

- Don't overwrite PYTHONPATH when running the unit tests, so the user's required packages located in PYTHONPATH directories are correctly loaded.

This solves one of the issues described in  #914

### DIFF
--- a/wscript
+++ b/wscript
@@ -339,7 +339,7 @@ def run_python_tests(ctx):
     os.system('cp -r src/python/essentia build/python/')
     os.system('cp build/src/python/_essentia*.so build/python/essentia')
 
-    ret = os.system('PYTHONPATH=build/python %s test/src/unittests/all_tests.py' % sys.executable)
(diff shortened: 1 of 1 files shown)


wscript
```diff
@@ -339,7 +339,7 @@ def run_python_tests(ctx):
     os.system('cp -r src/python/essentia build/python/')
     os.system('cp build/src/python/_essentia*.so build/python/essentia')
 
-    ret = os.system('PYTHONPATH=build/python %s test/src/unittests/all_tests.py' % sys.executable)
+    ret = os.system('PYTHONPATH=build/python:$PYTHONPATH %s test/src/unittests/all_tests.py' % sys.executable)
```
alastair, line 342: Maybe it'd be a good idea to use subprocess.run (python 3.5 and above), which takes a specific `env` argument, allowing us to run this without using a subshell. 
Double-check that this is valid if there is a space in `$PYTHONPATH`

PR #1010, opened 2020-07-04
Stop immediately if any command in a script fails

### LINKED ISSUE #986: Make 3rd party build scripts fail if any subcommand fails
We are using `build_3rdparty_static_debian.sh` in docker to build static dependencies. Sometimes, a part of one of these scripts fails, but then the rest of the build process continues. Docker thinks that everything is fine, because the scripts exited with a success code.

We should use bash's builtin `set -e` to cause these scripts to exit with an error if any subcommand fails. that way users can see immediately if some part of the process fails, and docker build will correctly stop if an error occurs.

amCap1712:
Fixes #986

### DIFF
--- a/packaging/build_3rdparty_static_debian.sh
+++ b/packaging/build_3rdparty_static_debian.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
 BASEDIR=$(dirname $0)
 cd $BASEDIR/debian_3rdparty
 ./build_eigen3.sh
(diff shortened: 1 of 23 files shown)


dbogdanov:
Thanks! As this is `bash` functionality we also need to change the shebang to `#!/usr/bin/env bash` in all the scripts updated in this PR.

We require [signing the Essentia and Gaia Individual Contributor License Agreement](https://essentia.upf.edu/contribute.html) for contributions larger than 20 lines of code. Can you take care of this before we merge your PRs?

amCap1712:
@dbogdanov I have emailed the CLA. I'll fix the shebangs by tomorrow.

amCap1712:
@dbogdanov Done.

dbogdanov:
Looks good now. Thanks!!

PR #1014, opened 2020-07-14
Several improvements related to  Eigen Tensors

### LINKED ISSUE #1013: Segmentation fault when accessing Essentia tensors from Python
Right now our `tensorReal` type is designed to serve Numpy array references (without coping the data) to Python.
https://github.com/MTG/essentia/blob/ba79be6515f2fd0cde75ee3f6fa98706a66f4c36/src/python/pytypes/tensorreal.cpp#L29

However, this scenario never happens as our only algorithm returning tensors (`PoolToTensor`) is only defined in streaming mode. Thus, in Python these tensors  have to be interfaced with `Pools`, where the tensor's data is copied before making the Numpy reference.
https://github.com/MTG/essentia/blob/ba79be6515f2fd0cde75ee3f6fa98706a66f4c36/src/python/pytypes/pypool.cpp#L534

Working on a standard algorithm returning a tensor, I have found memory deallocation errors when trying to access tensors' references in Python.

Given that we are already coping the tensors' data in the `Pools`, a solution would be to generalize this approach by changing `tensorReal`'s `toPythonRef` method to a `toPythonCopy` one.

pabloEntropia:
This PR contains several improvements related to  Eigen Tensors:

- TensorReal toPythonRef -> toPythonCopy to fix #1013 
- Add convenience aliases for tensors of different ranks
- Disable build-time Eigen-related warnings
- Add basic statistic operations with tensors (mean, std, min, max)
- Add new Algorithms `TensorTranspose` & `TensorNormalize`

### DIFF
--- a/src/algorithms/standard/tensornormalize.cpp
+++ b/src/algorithms/standard/tensornormalize.cpp
@@ -56,14 +56,34 @@ void TensorNormalize::compute() {
   switch (_scaler) {
     case STANDARD: {
       if (_axis == -1) {
-        Real means = mean(input);
-        Real stds = stddev(input, means);
+        Real globalMean = mean(input);
+        Real globalStd = stddev(input, globalMean);
 
-        output = (input - input.constant(means)) / input.constant(stds);
+        if (globalStd == 0) {
+          E_INFO("TensorNormalize: Received tensor with constant value.");
+          // Set std to 0 so we return a vector of all 0s.
+          globalStd = 1;
+        }
+
+        output = (input - globalMean) / globalStd;
 
       } else {
         Tensor<Real> means = mean(input, _axis);
         Tensor<Real> stds = stddev(input, means, _axis);
+
+        bool hasConstantSlice = false;
+
+        Real* stds_data = stds.data();
+        for (int i = 0; i < stds.size(); i++) {
+          if (stds_data[i] == 0) {
+            stds_data[i] = 1;
+            hasConstantSlice = true;
+          }
+        }
+
+        if (hasConstantSlice) {
+          E_INFO("TensorNormalize: Received tensor with constant value.");
+        }
 
         array<Eigen::Index, TENSORRANK> broadcastShape = input.dimensions();
         broadcastShape[_axis] = 1;
@@ -74,10 +94,16 @@ void TensorNormalize::compute() {
     }
     case MINMAX: {
       if (_axis == -1) {
-        Real minima = tensorMin(input);
-        Real maxima = tensorMax(input);
+        Real globalMinimum = tensorMin(input);
+        Real globalMaximum = tensorMax(input);
 
-        output = (input - minima) / (maxima - minima);
+        output = input - globalMinimum;
+
+        if (globalMaximum == globalMinimum) {
+          E_INFO("TensorNormalize: Received tensor with constant value.");
+        } else {
+          output /= output.constant(globalMaximum - globalMinimum);
+        }
 
       } else {
         Tensor<Real> minima = tensorMin(input, _axis);
@@ -86,8 +112,23 @@ void TensorNormalize::compute() {
         array<Eigen::Index, TENSORRANK> broadcastShape = input.dimensions();
         broadcastShape[_axis] = 1;
 
-        output = (input - minima.broadcast(broadcastShape)) /
-          (maxima - minima).broadcast(broadcastShape);
+        Tensor<Real> diff = maxima - minima;
+
+        bool hasConstantSlice = false;
+
+        Real* diff_data = diff.data();
+        for (int i = 0; i < diff.size(); i++) {
+          if (diff_data[i] == 0) {
+            diff_data[i] = 1;
+            hasConstantSlice = true;
+          }
+        }
+
+        if (hasConstantSlice) {
+          E_INFO("TensorNormalize: Received tensor with constant value.");
+        }
+
+        output = (input - minima.broadcast(broadcastShape)) / diff.broadcast(broadcastShape);
       }
       break;
     }
--- a/src/algorithms/standard/tensornormalize.cpp
+++ b/src/algorithms/standard/tensornormalize.cpp
@@ -61,7 +61,7 @@ void TensorNormalize::compute() {
 
         if (globalStd == 0) {
           E_INFO("TensorNormalize: Received tensor with constant value.");
-          // Set std to 0 so we return a vector of all 0s.
+          // Set std to 1 so we return a vector of all 0s.
           globalStd = 1;
         }
 
--- a/src/algorithms/standard/tensornormalize.cpp
+++ b/src/algorithms/standard/tensornormalize.cpp
@@ -1,0 +1,105 @@
+/*
+ * Copyright (C) 2006-2020  Music Technology Group - Universitat Pompeu Fabra
+ *
+ * This file is part of Essentia
+ *
+ * Essentia is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU Affero General Public License as published by the Free
+ * Software Foundation (FSF), either version 3 of the License, or (at your
+ * option) any later version.
+ *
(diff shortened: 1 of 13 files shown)


src/essentia/essentiamath.h
```diff
@@ -1200,6 +1200,142 @@ std::vector<std::vector<T> > pairwiseDistance(const std::vector<std::vector<T> >
...
 }
 
+/**
+ * Sets `squeezeShape`, `summerizerShape`, `broadcastShape` to perform operations 
+ * on a Tensor with the shape of `tesor` along the `axis` dimension.
+ */
+template <typename T>
+void tensorGeometricalInfo(const Tensor<T>& tensor, int& axis,
+                           std::array<Eigen::Index, TENSORRANK - 1>& squeezeShape,
+                           std::array<Eigen::Index, TENSORRANK>& summerizerShape,
```
dbogdanov, line 1210: "summarizerShare" (typo)

src/essentia/essentiamath.h
```diff
@@ -1200,6 +1200,142 @@ std::vector<std::vector<T> > pairwiseDistance(const std::vector<std::vector<T> >
   return pdist;
 }
 
+/**
+ * Sets `squeezeShape`, `summerizerShape`, `broadcastShape` to perform operations 
+ * on a Tensor with the shape of `tesor` along the `axis` dimension.
+ */
+template <typename T>
+void tensorGeometricalInfo(const Tensor<T>& tensor, int& axis,
```
dbogdanov, line 1208: Do we need to group computation for `squeezeShape`, `summarizerShape`and `broadcastShape` together in one method?

I guess it is fine if we use this mostly as an auxiliary method.

src/essentia/essentiamath.h
```diff
@@ -1200,6 +1200,142 @@ std::vector<std::vector<T> > pairwiseDistance(const std::vector<std::vector<T> >
   return pdist;
 }
 
+/**
+ * Sets `squeezeShape`, `summerizerShape`, `broadcastShape` to perform operations 
+ * on a Tensor with the shape of `tesor` along the `axis` dimension.
```
dbogdanov, line 1205: typo: `tesor` -> `tensor`

src/essentia/essentiamath.h
dbogdanov, line 1208: We can consider simplify the name: `Geometrical` -> `Geometric`. Also for consistency with other names (e.g., we have the `GeometricMean` algorithm)

test/src/unittests/standard/test_tensornormalize.py
```diff
@@ -0,0 +1,98 @@
...
+
+class TestTensorNormalize(TestCase):
+
+    def scalerSingleValue(self, scaler_arg=('standard', StandardScaler())):
+        scaler_name, scaler = scaler_arg
+
+        original = numpy.arange(1, dtype='float32')
+
+        # Use Scipy to generate the expected results
+        expected =scaler.fit_transform(original.reshape(-1, 1))
```
dbogdanov, line 34: " = "

test/src/unittests/standard/test_tensornormalize.py
```diff
@@ -0,0 +1,98 @@
...
+# You should have received a copy of the Affero GNU General Public License
+# version 3 along with this program. If not, see http://www.gnu.org/licenses/
+
+
+
+from essentia_test import *
+from sklearn.preprocessing import MinMaxScaler, StandardScaler
+
+
+class TestTensorNormalize(TestCase):
```
dbogdanov, line 26: What happens in the extreme cases (e.g., all value are the same)?

src/algorithms/standard/tensortranspose.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+const char* TensorTranspose::name = "TensorTranspose";
+const char* TensorTranspose::category = "Standard";
+const char* TensorTranspose::description = DOC("This algorithm performs transpositions over the axes of a tensor.\n");
+
+  void TensorTranspose::configure() {
+    if (!parameter("permutation").isConfigured()) return;
+    _permutation = parameter("permutation").toVectorInt();
+
+    if (_permutation.size() != TENSORRANK) {
```
dbogdanov, line 36: These checks do not take into consideration the case of  having the same value repeated in the permutation vector.

src/algorithms/standard/tensornormalize.h
```diff
@@ -0,0 +1,92 @@
...
+  int _axis;
+
+ public:
+  TensorNormalize() {
+    declareInput(_input, "tensor", "the input tensor");
+    declareOutput(_output, "tensor", "the normalized output tensor");
+  }
+
+  void declareParameters() {
+    declareParameter("scaler", "the type of the normalization to apply to input tensor", "{standard,minMax}", "standard");
```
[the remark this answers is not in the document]
pabloEntropia: I borrowed the names from scikit-learn methods:
 - [StandardScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html)
- [MinMaxScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html)

Also, in the algorithm documentation I refer to [this](https://en.wikipedia.org/wiki/Feature_scaling) Wikipedia article that calls it `Standardization`.

About the parameter values, I have found 14 parameters using values with underscore and 9 using lower camel case.
Interestingly, some algorithms like `MelBands` and `MFCC` have parameters with both styles.

{slaneyMel,htkMel} -> MelBands, MFCC
{none,unitSum,unitMax} -> HPCP
{none,cosine,squaredCosine} -> KeyExtractor, HPCP
{linear,slaneyMel,htkMel} -> TriangularBands
{sumEnergy, peakEnergy} -> SingleBeatLoudness

{none,unit_sum,unit_max} -> Chromagram, Histogram
{starting_beat,interbeat_median} -> ChordDetectionBeats
{unit_sum,unit_tri,unit_max} -> MelBands, MFCC, TriangularBands
{unit_sum,unit_max} -> BFCC, SpectrumToCent, TriangularBarkBands
{flat,triangle,inverse_triangle,parabola,inverse_parabola,linear,quadratic,inverse_quadratic,hybrid,supplied} -> NoveltyCurve
{infogain,beat_emphasis} -> OnsetDetectionGlobal
{hfc,complex,complex_phase,flux,melflux,rms} -> OnsetDetection

We should decice something about this. I would prefer lower camel case for consistency with the parameter names, although underscores seem more pythonic.

src/essentia/essentiamath.h
pabloEntropia: Yes, I thought it could be a useful auxiliary method in case we need to implement more axis-wise tensor operations in the future.

pabloEntropia: I used `geometrical`  for consistency with  Eigen [Geometrical operations ](https://eigen.tuxfamily.org/dox/unsupported/eigen_tensors.html#autotoc_md95), but I do not have a preference for one name or another.

src/algorithms/standard/tensornormalize.cpp
```diff
@@ -56,15 +56,35 @@ void TensorNormalize::compute() {
...
       if (_axis == -1) {
-        Real means = mean(input);
-        Real stds = stddev(input, means);
+        Real globalMean = mean(input);
+        Real globalStd = stddev(input, globalMean);
 
-        output = (input - input.constant(means)) / input.constant(stds);
+        if (globalStd == 0) {
+          E_INFO("TensorNormalize: Received tensor with constant value.");
+          // Set std to 0 so we return a vector of all 0s.
```
dbogdanov, line 64: Typo here? (should be "Set std to 1...")?

src/essentia/essentiamath.h
dbogdanov: Then, this is fine

PR #1017, opened 2020-07-22
melspectrogram and batch_melspectrogram

pabloEntropia:
Some Pytools changes and utility functions.
- Batch extractors are moved from `extractors.py` to `batch_extractors.py`. `extractors` is now a package with separated modules for each extractor (so far `music_extractor` and `melspectrogram`) 
- Add `melspectrogram` extractor that can be used as a module or as a script.
- Add `batch_melspectrogram` that relies on `melspetrogram` to analyze entire folder structures.
- Both functions support most of the algorithms' parameters


src/python/essentia/pytools/extractors/melspectrogram.py
```diff
@@ -0,0 +1,227 @@
...
+LOW_FREQUENCY_BOUND = 0.
+HIGH_FREQUENCY_BOUND = 22050.
+WARPING_FORMULA = 'htkMel'
+WEIGHTING = 'warping'
+NORMALIZE = 'unit_sum'
+BANDS_TYPE = 'power'
+COMPRESSION_TYPE = 'shift_scale_log'
+
+
+def melspectrogram(filename, npy_file=None, force=False, verbose=False, sample_rate=SAMPLE_RATE, frame_size=FRAME_SIZE,
```
dbogdanov, line 44: Note that it is possible to use `**kwargs`  for non-essential parameters like here: https://github.com/MTG/essentia/blob/master/src/python/essentia/pytools/spectral.py#L250

dbogdanov:
Have a look at 
https://github.com/google/python-fire
http://micheles.github.io/plac/

src/python/essentia/pytools/extractors/melspectrogram.py
pabloEntropia: Right, the function definition looks ugly with so many parameters. However, I decided to explicitly mention all the parameters because I wanted `help(melspectrogram)` to show all the parameters available with their default values. Especially because some parameters (e.g., `compression_type`) doesn't match any parameter from Essentia algorithms directly. 
Do you think there is a better way to achieve this?

```
melspectrogram?
```

```ShellSession
Signature:
melspectrogram(
    filename,
    npy_file=None,
    force=False,
    verbose=False,
    sample_rate=44100.0,
    frame_size=1024,
    hop_size=512,
    window_type='hann',
    zero_padding=0,
    low_frequency_bound=0.0,
    high_frequency_bound=22050.0,
    number_bands=24,
    warping_formula='htkMel',
    weighting='warping',
    normalize='unit_sum',
    bands_type='power',
    compression_type='shift_scale_log',
)
Docstring:
Computes the mel spectrogram given the audio filename.
When the parameter `npy_file` is specified, the data is saved to disk as a numpy array (.npy).
Use the parameter `force` to overwrite the numpy array in case it already exists.
The rest of parameters are directly mapped to Essentia algorithms as explained below.
...
```

pabloEntropia:
> Have a look at
> https://github.com/google/python-fire
> http://micheles.github.io/plac/

Both look like good alternatives. Maybe they could help to make it less verbose at the expense of having another dependency to Essentia.
Also notice that the current solution allows to have different behaviors for the module and the script behaviors. For instance, `npy_file` is a required argument in the script behavior but it's optional when called as a module. I belive this wouldn't be posible with `fire`.

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+from multiprocessing import Pool
+from multiprocessing import cpu_count
+from subprocess import Popen, PIPE
+from essentia import EssentiaError
+from functools import partial
+import os
+import sys
+
+
+def __subprocess(cmd, verbose=True):
```
alastair, line 28: I would name this with a single _, normally  2 means that it's internal to python.

the subprocess module has `run` since v 3.5, perhaps we could use this to reduce some of this code? https://docs.python.org/3.7/library/subprocess.html#using-the-subprocess-module

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+    return rc, cmd_str, stderr.decode("utf-8")
+
+
+def __batch_extractor(audio_dir, output_dir, extractor_cmd, output_extension,
+                      generate_log=True, audio_types=None, skip_analyzed=False,
+                      jobs=0, verbose=True):
+    if not audio_types:
+        audio_types = ('.wav', '.aiff', '.flac', '.mp3', '.ogg')
+        print("Audio files extensions considered by default: " +
+              ' '.join(audio_types))
```
alastair, line 55: personal preference here, I'd probably separate these by a ,
do you want to consider using a logger instead of a print in case people want to turn it off?

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+                  'the analysis will try to continue with 4 jobs. '
+                  'For a different behavior change the `job` parameter.')
+            jobs = 4
+
+    skipped_count = 0
+    skipped_files = []
+    cmd_lines = []
+    for root, _, filenames in os.walk(audio_dir):
+        for filename in filenames:
+            if filename.lower().endswith(audio_types):
```
alastair, line 82: wow, I didn't know this feature! interesting!
I've also done it by doing `if os.path.splitext(filename.lower())[1] in audio_types`

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+                if skip_analyzed:
+                    if os.path.isfile( '{}.{}'.format(out_file, output_extension)):
+                        print("Found descriptor file for " +
+                              audio_file + ", skipping...")
+                        skipped_files.append(audio_file)
+                        skipped_count += 1
+                        continue
+                folder = os.path.dirname(out_file)
+                if not os.path.exists(folder):
+                    os.makedirs(folder)
```
alastair, line 95: makedirs has exist_ok=True

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+
+                elif os.path.isfile(folder):
```
alastair, line 97: Is this a typically error that people might get? I would just typically do a check if it exists, regardless of if it's a file or a directory, so just a simple `os.path.exists`, and raise an error if necessary?

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+                    raise EssentiaError('Cannot create directory {}. '
+                                        'There exist a file with the same name. '
+                                        'Aborting analysis.'.format(folder))
+
+                cmd_lines.append(extractor_cmd + [audio_file, out_file])
+
+    # analyze
+    log_lines = []
+    total, errors, oks =0, 0, 0
```
alastair, line 106:
```suggestion
    total, errors, oks = 0, 0, 0
```

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+    if len(cmd_lines) > 0:
```
alastair, line 107:
```suggestion
    if cmd_lines:
```

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+        p = Pool(jobs)
```
alastair, line 108: can we use concurrent futures here?

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+                oks += 1
+                log_lines.append('"{}" ok!'.format(cmd_idx))
+            else:
+                errors += 1
+                log_lines.append('"{}" failed'.format(cmd_idx))
+                log_lines.append('  "{}"'.format(err))
+
+    summary = ("Analysis done for {} files. {} files have been skipped due to errors, "
+               "{} were successfully processed and {} already existed.\n").format(total, errors, oks, skipped_count)
+    print(summary)
```
alastair, line 126: do you want the summary information to be machine readable or human readable?

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+                         weighting=None, normalize=None, bands_type=None, compression_type=None):
+    """Generates mel bands for every audio file matching `audio_types` in `audio_dir`.
+    The generated .npy files are stored in `output_dir` matching the folder
+    structure found in `audio_dir`.
+    """
+
+    extractor_cmd = [sys.executable, os.path.join(os.path.dirname(__file__),
+                                                  'extractors/melspectrogram.py')]
+
+    # Set --force as a hardcoded flat.
```
alastair, line 171: flag?

src/python/essentia/pytools/extractors/melspectrogram.py
```diff
@@ -0,0 +1,229 @@
...
+    hop_size=HOP_SIZE, window_type=WINDOW_TYPE, zero_padding=ZERO_PADDING, low_frequency_bound=LOW_FREQUENCY_BOUND,
+    high_frequency_bound=HIGH_FREQUENCY_BOUND, number_bands=NUMBER_BANDS, warping_formula=WARPING_FORMULA,
+    weighting=WEIGHTING, normalize=NORMALIZE, bands_type=BANDS_TYPE, compression_type=COMPRESSION_TYPE):
+    """Computes the mel spectrogram given the audio filename.
+    When the parameter `npy_file` is specified, the data is saved to disk as a numpy array (.npy).
+    Use the parameter `force` to overwrite the numpy array in case it already exists.
+    The rest of parameters are directly mapped to Essentia algorithms as explained below.
+
+    Note: this functionallity is also avaiable as a command line script.
```
alastair, line 53:
```suggestion
    Note: this functionality is also available as a command line script.
```

Do you also want to explain how to call this commandline script? Where will people be seeing this comment?

src/python/essentia/pytools/extractors/melspectrogram.py
```diff
@@ -0,0 +1,229 @@
...
+        
+    if verbose:
+        print('Done for "{}"'.format(npy_file))
+
+    return mel_bands
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Computes the mel spectrogram given the audio filename.')
```
alastair, line 191:
```suggestion
        description='Computes the mel spectrogram of a given audio file.')
```

src/python/essentia/pytools/extractors/melspectrogram.py
```diff
@@ -0,0 +1,229 @@
...
+    Returns:
+        (2D array): The mel-spectrogram.
+    """
+
+    padded_size = frame_size + zero_padding
+    spectrum_size = (padded_size) // 2 + 1
+
+
+    # In case we want to save the melbands to a file
+    # check if the file already existss
```
alastair, line 124:
```suggestion
    # check if the file already exists
```

src/python/essentia/pytools/extractors/melspectrogram.py
alastair: agree that keeping the explicit args for the documentation is a good idea

src/python/essentia/pytools/batch_extractors.py
```diff
@@ -0,0 +1,219 @@
...
+
+    summary = ("Analysis done for {} files. {} files have been skipped due to errors, "
+               "{} were successfully processed and {} already existed.\n").format(total, errors, oks, skipped_count)
+    print(summary)
+
+    # generate log
+    if generate_log:
+        log = [summary] + log_lines
+
+    with open(os.path.join(output_dir, 'log'), 'w') as f:
```
alastair, line 132: should this be inside `if generate_log`?

src/python/essentia/pytools/batch_extractors.py
pabloEntropia: We just want to raise an error if it is a file. For the folder case, it can be the result of a previous incomplete analysis that the user wants to resume.

src/python/essentia/pytools/batch_extractors.py
pabloEntropia: Sounds good.
Maybe we should make a separate PR moving from print to logger all the `pytools`.

dbogdanov: Yes, this should be a separate PR.

dbogdanov (approved):
Looks good!

PR #1033, opened 2020-08-27
PercivalBpmEstimator: push float literal instead of double

drlight-code:
fix error due to missing conversion from double to Real. gcc 10.2.0
RuntimeError: In PercivalBpmEstimator.compute: While trying to push item into source PercivalBpmEstimator::bpm:
Error when checking types. Expected: Real, received: double


dbogdanov:
Thanks for noticing this!

PR #1042, opened 2020-09-01
Many dependencies required on Ubuntu 18.04

hobson:
./waf --configure will fail if `build_debian.sh` is not first run


alastair:
Hi, thanks for the PR!

We have instructions to install dependencies on the essentia website: https://essentia.upf.edu/installing.html#installing-dependencies-on-linux
This script that you included appears to be a helper for a very specific task (building the music extractor), and so doesn't fit well as a general-purpose helper tool.
If you still think that there is a good reason to keep this script then please let us know your reasoning behind it.

PR #1057, opened 2020-10-09
add sphinx doc tag to fix link.

idrisr:
Currently the link leads to a dead page.


dbogdanov:
Hi @idrisr, thanks for noticing that!

PR #1058, opened 2020-10-09
modify script to work on mac as well as linux

idrisr:
The command `./waf doc` doesn't work on Mac OS X, as it complains it can't find the directory. Using the `.` fixes that issue. I've tested this on Mac OS X 10.15.4 and on Ubuntu 20.04 and both worked as expected.


dbogdanov:
Cannot reproduce this issue on macOS 11.3.1, so I am closing this for now.

Issue #835, opened 2019-03-06
memory leak in `multipitchmelodia`


beantowel:
repeatedly use MultipitchMelodia in python will cause the memory usage grow rapidly and result in std::badalloc

example code:
```
elFilter = EqualLoudness(sampleRate=sr)
pitchEst = MultiPitchMelodia(sampleRate=sr)
pitchCandidates = pitchEst(signal)
```

dbogdanov:
Please, report more details (Essentia version, OS, how do you run the code repeatedly). 

Having done a quick test, I cannot reproduce this issue on my Ubuntu 18.10 machine with the latest build from the master branch.

beantowel:
sorry, i was too hurry to check if the simplified code would reproduce the error. I'll attach the detailed code that fails on my machine, the status of which was listed below:
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.2 LTS
Release:	18.04
Codename:	bionic

[reproduceMMLeak.tar.gz](https://github.com/MTG/essentia/files/2936695/reproduceMMLeak.tar.gz)

$ uname -r
4.15.0-46-lowlatency

$ pip show essentia
Name: essentia
Version: 2.1b5.dev532
Summary: Library for audio and music analysis, description and synthesis
Home-page: http://essentia.upf.edu
Author: Dmitry Bogdanov
Author-email: dmitry.bogdanov@upf.edu
License: AGPLv3
Location: /home/beantowel/anaconda3/lib/python3.6/site-packages
Requires: numpy, pyyaml, six
Required-by:
```
[reproduceMMLeak.tar.gz](https://github.com/MTG/essentia/files/2936699/reproduceMMLeak.tar.gz)

cifkao:
Could #1054 be the real problem (or a related one)? @beantowel @dbogdanov

PR #1059, opened 2020-10-10
Fix infinite loop in PitchContoursMultiMelody

### LINKED ISSUE #1054: Infinite loop in MultiPitchMelodia.compute?
`MultiPitchMelodia` (used from Python) never finishes for some inputs. Here is an example input that triggers this: [badfile.zip](https://github.com/MTG/essentia/files/5322841/badfile.zip)

Code:
```python
import essentia.standard as es
audio = es.MonoLoader(filename='badfile.wav')()
pitches = es.MultiPitchMelodia()(audio)
```

**Edit:** The code does finish after a few minutes by raising a `RuntimeError: In MultiPitchMelodia.compute: std::bad_alloc`.

Adding some noise to the audio (e.g. `np.random.normal(loc=audio, scale=1e-4)`) fixes the problem for this particular file.

I have essentia 2.1-beta6-dev.

cifkao:
Fixes #1054; #835 may be related too. There is a loop that keeps adding new elements to `pitches` instead of updating the existing ones, therefore the loop never ends and the vector eventually fills up the memory.

### DIFF
--- a/src/algorithms/tonal/pitchcontoursmultimelody.cpp
+++ b/src/algorithms/tonal/pitchcontoursmultimelody.cpp
@@ -142,10 +142,8 @@ void PitchContoursMultiMelody::compute() {
 
   // no contours -> zero pitch vector output
   if (contoursBins.empty()) {
-      vector<Real> zero;
-      zero.push_back(0.0);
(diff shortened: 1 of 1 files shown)


dbogdanov:
Thank you!

PR #1084, opened 2020-12-04
Request to add additions to the onsetdetection function

cvf-bcn-gituser:
-onset detection functions used in assertEqual asserts 
-assertEqualVector tests corrrections
-redundant commented out code removed

### DIFF
--- a/test/src/unittests/rhythm/test_onsetdetection.py
+++ b/test/src/unittests/rhythm/test_onsetdetection.py
@@ -36,6 +36,10 @@ class TestOnsetDetection(TestCase):
         fft = FFT()
         onset_hfc = OnsetDetection(method='hfc')
         onset_complex = OnsetDetection(method='complex')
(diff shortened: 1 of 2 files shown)


test/src/unittests/rhythm/test_onsetdetection.py
```diff
@@ -128,12 +151,14 @@ def testEmpty(self):
         self.assertComputeFails(OnsetDetection(), spectrum, phase)
         spectrum = ones(1024)
         self.assertComputeFails(OnsetDetection(method='complex'), spectrum, phase)
-
-
+        self.assertComputeFails(OnsetDetection(method='complex_phase'), spectrum, phase)
+    
     def testDifferentSizes(self):
```
dbogdanov, line 156: Update the name of this test to represent that it is only for "complex*" tests. For example, `testComplexInputSizeMismatch` is more explanatory.

dbogdanov (changes requested):

PR #1103, opened 2021-02-08
Updates to superflux extractor (e.g RegTest, invalid vals, etc.)

cvf-bcn-gituser:
Added Regression Tests based on two audio files, whose values were captured on previous runs with the same parameters.
Tested on Ubuntu and Linux Subsystem on Windows.
Superfluxextractor covers peaks and novelty tests internally.
Dummies added  to maintain 1:1 correspondence between src, file  and UT.


test/src/unittests/rhythm/test_superfluxextractor.py
```diff
@@ -17,12 +17,60 @@
...
+            for j in range(1,len(expected)):
+                if found[i] <= expected[j] and found[i] >= expected[j-1]:
+                    if fabs(found[i] - expected[j-1]) < fabs(expected[j] - found[i]):
+                        self.assertAlmostEqual(found[i], expected[j-1], precision)
+                    else:
+                        self.assertAlmostEqual(found[i], expected[j], precision)
+
     def testSilence(self):
         # zeros should return no onsets (empty array)
         self.assertEqualMatrix(SuperFluxExtractor()(zeros(44100)), [])
```
dbogdanov, line 76: The output is 1-D vector, so we can use `assertEqualVector` instead.

test/src/unittests/rhythm/test_superfluxextractor.py
```diff
@@ -55,7 +103,6 @@ def testImpulse(self):
         precission = (hopSize + frameSize) / sampleRate
```
dbogdanov, line 103: MIsspel: `precision`

dbogdanov (changes requested):
There's a redundant call `self.assertComputeFails(Onsets(), array([[]]), [])`, possibly left from another test.

test/src/unittests/rhythm/test_superfluxextractor.py
cvf-bcn-gituser: self.assertEqualMatrix should use self.assertEqualVector

PR #1104, opened 2021-02-11
Unit tests for LoopBpmEstimator and LoopBpmConfidence

cvf-bcn-gituser:
During Unit Tests a minor issue was found was found in loopbpmconfidence.cpp, when dealing with empty incoming signals. A fix was added for this.

The "techno_loop.wav'" recording forms the basis of the tests, using the whole signal and parts of the signal for different checks.

### DIFF
--- a/src/algorithms/rhythm/loopbpmconfidence.cpp
+++ b/src/algorithms/rhythm/loopbpmconfidence.cpp
@@ -54,66 +54,71 @@ void LoopBpmConfidence::compute() {
   } else {
     const vector<Real>& signal = _signal.get();
 
-    // Get original duration
+    // Get original duration.
     int duration_samples = signal.size();
 
-    // Compute envelope
-    vector<Real> envelope;
-    _envelope->input("signal").set(signal);
-    _envelope->output("signal").set(envelope);
-    _envelope->compute();
+    // Check first that the signal is non-empty
+    if (duration_samples!=0){
+      // Compute envelope
+      vector<Real> envelope;
+      _envelope->input("signal").set(signal);
+      _envelope->output("signal").set(envelope);
+      _envelope->compute();
 
-    // Compute threshold
-    Real threshold = *std::max_element(envelope.begin(), envelope.end()) * 0.05;
+      // Compute threshold
+      Real threshold = *std::max_element(envelope.begin(), envelope.end()) * 0.05;
 
-    // Find start position
-    int start_position = 0;
-    for (int i=0; i<(int)envelope.size(); i++){
-      if (envelope[i] >= threshold){
-        start_position = i;
-        break;
-      }
-    }
-
-    // Find end position
-    int end_position = 0;
-    for (int i=envelope.size() - 1; i>=0; i--){
-      if (envelope[i] >= threshold){
-        end_position = i;
-        break;
-      }
-    }
-
-    // Build vector with all durations to check
-    std::vector<int> durations_to_check;
-    durations_to_check.resize(4);
-    durations_to_check[0] = duration_samples;
-    durations_to_check[1] = duration_samples - start_position;
-    durations_to_check[2] = end_position;
-    durations_to_check[3] = end_position - start_position;
-
-    // Check all durations
-    std::vector<Real> confidences;
-    confidences.resize(4);
-    Real beatDuration = (60.0 * parameter("sampleRate").toReal()) / bpmEstimate;
-    Real lambdaThreshold = beatDuration * 0.5;
-    for (int i=0; i<(int)durations_to_check.size(); i++){
-      int duration = durations_to_check[i];
-      int minDistance = duration_samples; // Set maximum duration
-      for (int j=1; j<128; j++) { // Iterate over possible beat lengths (1-127)
-        int nBeatDuration = (int)round(beatDuration * j);
-        int distance = abs(duration - nBeatDuration);
-        if (distance < minDistance) {
-          minDistance = distance;
+      // Find start position
+      int start_position = 0;
+      for (int i=0; i<(int)envelope.size(); i++){
+        if (envelope[i] >= threshold){
+          start_position = i;
+          break;
         }
       }
-      if (minDistance > lambdaThreshold) {
-        confidences[i] = 0.0;
-      } else {
-        confidences[i] = 1.0 - (Real)minDistance / lambdaThreshold;
+
+      // Find end position
+      int end_position = 0;
+      for (int i=envelope.size() - 1; i>=0; i--){
+        if (envelope[i] >= threshold){
+          end_position = i;
+          break;
+        }
       }
-    }  
-    confidence = *std::max_element(confidences.begin(), confidences.end());
+
+      // Build vector with all durations to check
+      std::vector<int> durations_to_check;
+      durations_to_check.resize(4);
+      durations_to_check[0] = duration_samples;
+      durations_to_check[1] = duration_samples - start_position;
+      durations_to_check[2] = end_position;
+      durations_to_check[3] = end_position - start_position;
+
+      // Check all durations
+      std::vector<Real> confidences;
+      confidences.resize(4);
+      Real beatDuration = (60.0 * parameter("sampleRate").toReal()) / bpmEstimate;
+      Real lambdaThreshold = beatDuration * 0.5;
+      for (int i=0; i<(int)durations_to_check.size(); i++){
+        int duration = durations_to_check[i];
+        int minDistance = duration_samples; // Set maximum duration
+        for (int j=1; j<128; j++) { // Iterate over possible beat lengths (1-127)
+          int nBeatDuration = (int)round(beatDuration * j);
+          int distance = abs(duration - nBeatDuration);
+          if (distance < minDistance) {
+            minDistance = distance;
+          }
+        }
+        if (minDistance > lambdaThreshold) {
+          confidences[i] = 0.0;
+        } else {
+          confidences[i] = 1.0 - (Real)minDistance / lambdaThreshold;
+        }
+      }  
+      confidence = *std::max_element(confidences.begin(), confidences.end());
+    } else {
+      confidence = 0.0;  // Confidence set to zero for empty signals.
+    }
   }
 }
 
--- a/test/src/unittests/rhythm/test_loopbpmconfidence.py
+++ b/test/src/unittests/rhythm/test_loopbpmconfidence.py
@@ -135,34 +135,27 @@ class TestLoopBpmConfidence(TestCase):
     def testSilentEdges(self):
 
         audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))() 
-        bpmEstimate = 125
+        bpmEstimate = 120
         confidence = LoopBpmConfidence()(audio, bpmEstimate)              
         lenSilence = int(0.01*len(audio)) # Choose a silent length 1% of audio
         silentAudio = zeros(lenSilence)
+        benchmarkConfidence= 0.89  # This figure was arrived at emperically from the min. confidence observed with test runs 
 
         # case 1: there is non-musical* silence before the loop starts
         signal1  = numpy.append(silentAudio, audio)
         confidence = LoopBpmConfidence()(signal1, bpmEstimate)              
-        self.assertGreater(confidence, 0.9) 
-
-        # Silent Edges dont always work.
-        # When adding silence to the end of this loop the confidence drops.
-        # This isnt a big problem since it is just a way 
-        # to try to better estimate confidence if the loop is not well cut.
-        # In any case, the "mellifluousanonymous" loop mentioned above
-        # works well with silence added at the end.
+        self.assertGreater(confidence,benchmarkConfidence)
 
         # case 2: there is non-musical silence after the loop ends        
         signal2  = numpy.append(audio,silentAudio)
         confidence = LoopBpmConfidence()(signal2, bpmEstimate)  
-        self.assertGreater(confidence, 0.27) 
-
+        self.assertGreater(confidence, benchmarkConfidence) 
 
         # case 3: there is non-musical silence at both ends
         signal3 = numpy.append(signal1, silentAudio)
         confidence = LoopBpmConfidence()(signal3, bpmEstimate)              
-        self.assertGreater(confidence, 0.27) 
-          
+        self.assertGreater(confidence, benchmarkConfidence) 
+
     def testEmpty(self):
         # Zero estimate check results in zero confidence
         emptyAudio = []
--- a/test/src/unittests/rhythm/test_loopbpmconfidence.py
+++ b/test/src/unittests/rhythm/test_loopbpmconfidence.py
@@ -29,31 +29,47 @@ class TestLoopBpmConfidence(TestCase):
 
     def testRegression(self):
         audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))()        
+
+        # Test for different BPMs starting with correct one then with decreasing accuracy
+        # and corresponding levels decreasing confidence.
+        # Figures obtained from previous runs algorithm as of baseline 10-Feb-2021
+        # In keeping with regression tests principles , no future changes of algorothm
+        # should break these tests.
+        bpmEstimate = 125
+        expectedConfidence = 1
+        confidence = LoopBpmConfidence()(audio,bpmEstimate)        
+        self.assertAlmostEqual(expectedConfidence, confidence, 0.1)
+
         bpmEstimate = 120
         expectedConfidence = 0.87
-        confidence = LoopBpmConfidence()(audio,bpmEstimate)        
-        self.assertAlmostEqual( expectedConfidence, confidence,0.1)
+        confidence = LoopBpmConfidence()(audio, bpmEstimate)        
+        self.assertAlmostEqual(expectedConfidence, confidence, 0.1)
+
+        bpmEstimate = 70
+        expectedConfidence = 0.68
+        confidence = LoopBpmConfidence()(audio, bpmEstimate)                
+        self.assertAlmostEqual(expectedConfidence, confidence, 0.1)
 
     def testEmpty(self):
         emptyAudio = []
         bpmEstimate = 0
         expectedConfidence = 0
-        confidence = LoopBpmConfidence()(emptyAudio,bpmEstimate)                  
-        self.assertEquals( expectedConfidence, confidence)
+        confidence = LoopBpmConfidence()(emptyAudio, bpmEstimate)                  
+        self.assertEquals(expectedConfidence, confidence)
 
     def testZero(self):
         zeroAudio = zeros(1000)
         bpmEstimate = 0
         expectedConfidence = 0
-        confidence = LoopBpmConfidence()(zeroAudio,bpmEstimate)
-        self.assertEquals( expectedConfidence, confidence)
+        confidence = LoopBpmConfidence()(zeroAudio, bpmEstimate)
+        self.assertEquals(expectedConfidence, confidence)
 
     def testConstantInput(self):
         onesAudio = ones(100)
         bpmEstimate = 0
         expectedConfidence = 0
-        confidence = LoopBpmConfidence()(onesAudio,bpmEstimate)
-        self.assertEquals( expectedConfidence, confidence)
+        confidence = LoopBpmConfidence()(onesAudio, bpmEstimate)
+        self.assertEquals(expectedConfidence, confidence)
 
 suite = allTests(TestLoopBpmConfidence)
 
--- a/test/src/unittests/rhythm/test_loopbpmconfidence.py
+++ b/test/src/unittests/rhythm/test_loopbpmconfidence.py
@@ -132,29 +132,37 @@ class TestLoopBpmConfidence(TestCase):
     Back end silence Confidence  is  0.9915646314620972
     Silence Both end confidence is  0.9915646314620972    
     """
-    def testSilentEdges(self):
-
+    def testSilentEdge(self):
         audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))() 
-        bpmEstimate = 120
+        bpmEstimate = 125
         confidence = LoopBpmConfidence()(audio, bpmEstimate)              
-        lenSilence = int(0.01*len(audio)) # Choose a silent length 1% of audio
+        lenSilence = 30000 # N.B The beat period is 21168 samples for 125 bpm @ 44.1k samp. rate
         silentAudio = zeros(lenSilence)
-        benchmarkConfidence= 0.89  # This figure was arrived at emperically from the min. confidence observed with test runs 
+        benchmarkConfidence = 0.96 # This figure was arrived at emperically from the min. confidence observed with test runs 
 
-        # case 1: there is non-musical* silence before the loop starts
+        # Test addition of non-musical silence before the loop starts
+        # The length is not a beat period,
+        # Nonetheless, we can stillreliably estimate the starting point because it is a hard transient.
         signal1  = numpy.append(silentAudio, audio)
-        confidence = LoopBpmConfidence()(signal1, bpmEstimate)              
-        self.assertGreater(confidence,benchmarkConfidence)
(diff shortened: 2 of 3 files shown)


test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,157 @@
...
+        
+        bpmEstimate = 70
+        expectedConfidence = 0.74
+        confidence = LoopBpmConfidence()(audio[samplePoint1:samplePoint2], bpmEstimate)            
+        self.assertAlmostEqual(expectedConfidence, confidence, 0.02) 
+
+        # Randomly chosen points, Section 2
+        samplePoint1 = 22000
+        samplePoint2 = 38000
+        expectedEstimate = 100
```
dbogdanov, line 100: This accounts for 0.36 seconds. This is too short (see the previous comment) as we want something that is long enough to account for a large segment of a 4/4 bar.

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,157 @@
...
+        self.assertEquals(0, confidence)
+
+    def testZero(self):
+        zeroAudio = zeros(1000)
+        bpmEstimate = 0
+        confidence = LoopBpmConfidence()(zeroAudio, bpmEstimate)
+        self.assertEquals(0, confidence)
+
+        # Non-zero estimate check results in non-zero confidence
+        # FIX-ME: Maybe given contant zero input, we should expect zero confidence
```
dbogdanov, line 137: We use `FIXME` for consistency.

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,157 @@
...
+        bpmEstimate = 125
+        confidence = LoopBpmConfidence()(zeroAudio, bpmEstimate)
+        self.assertNotEquals(0, confidence)
```
dbogdanov, line 140: @ffont Can you comment your opinion on this?

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,157 @@
...
+        onesAudio = ones(100)
+        bpmEstimate = 0
+        confidence = LoopBpmConfidence()(onesAudio, bpmEstimate)
+        self.assertEquals(0, confidence)
+
+        # Non-zero estimate check results in non-zero confidence
+        # FIX-ME: Maybe given contant input, we should expect zero confidence
+        bpmEstimate = 125
+        confidence = LoopBpmConfidence()(onesAudio, bpmEstimate)
+        self.assertNotEquals(0, confidence)
```
dbogdanov, line 152: @ffont What is the desired output in this case?

test/src/unittests/rhythm/test_loopbpmestimator.py
```diff
@@ -0,0 +1,120 @@
...
+
+class TestLoopBpmEstimator(TestCase):
+
+    def testInvalidParam(self):
+        self.assertConfigureFails(LoopBpmEstimator(), {'confidenceThreshold': -1})
+
+    def testRegression(self):
+        audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))()        
+      
+        expectedEstimate= 125
```
dbogdanov, line 33: `expectedEstimate = 125` (spacing)

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,157 @@
...
+        self.assertEquals(0, confidence)
+
+        # Non-zero estimate check results in non-zero confidence
+        # FIX-ME: Maybe given contant zero input, we should expect zero confidence
+        bpmEstimate = 125
+        confidence = LoopBpmConfidence()(zeroAudio, bpmEstimate)
+        self.assertNotEquals(0, confidence)
+         
+    def testConstantInput(self):
+        onesAudio = ones(100)
```
dbogdanov, line 143: Let's use the audio length consistent with testZero. Also, in both tests, 1000 samples is a too small value for BPM estimation. Let's set it to something meaningful like 2 seconds (which is a full bar in 120 BPM)

test/src/unittests/rhythm/test_loopbpmestimator.py
```diff
@@ -0,0 +1,120 @@
...
+
+        # Further integrity checks at randomly chosen points in audio.
+        # If any future changes break these asserts,
+        # then this will indicates algorithm is compromised.
+
+        # Randomly chosen points, Section 1
+        # A segment of techno_loop gives a non-ideal estimation of BPM.
+        samplePoint1 = 1000
+        samplePoint2 = 6000
+        expectedEstimate = 100.0
```
dbogdanov, line 53: See the previous comments. We can have a larger sub-segment.

test/src/unittests/rhythm/test_loopbpmestimator.py
```diff
@@ -0,0 +1,120 @@
...
+
+        # High confidence expectation triggers 0 bpm result in this section 1
+        estimate = LoopBpmEstimator(confidenceThreshold=0.95)(audio[samplePoint1:samplePoint2])            
+        self.assertEqual(0, estimate) 
+
+        # Randomly chosen points, Section 2
+        # A segment of techno_loop gives a non-ideal estimation of BPM.        
+        samplePoint1 = 22000
+        samplePoint2 = 38000
+        expectedEstimate = 100.0
```
dbogdanov, line 79: Same here. 

We can use the same sub-segments as in the LoopBpmConfidence algorithm.

test/src/unittests/rhythm/test_loopbpmestimator.py
```diff
@@ -0,0 +1,120 @@
...
+        emptyAudio = []
+        self.assertRaises(RuntimeError, lambda: LoopBpmEstimator()(emptyAudio))
+
+    def testZero(self):
+        zeroAudio = zeros(1000)
+        self.assertRaises(RuntimeError, lambda: LoopBpmEstimator()(zeroAudio))
+
+    def testConstantInput(self):
+        onesAudio = ones(100)
+        self.assertRaises(RuntimeError, lambda: LoopBpmEstimator()(onesAudio))
```
dbogdanov, line 114: @ffont Is this the desired behavior for LoopBpmEstimator? Specifically, if we were to use this in the Freesound extractor.

dbogdanov: Maybe we should always give a value on a legit audio input, including silence or constant input. This will be also consistent with RhythmExtractor2013.

dbogdanov (changes requested):

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,157 @@
...
+         should break these tests.
+         Test for different BPMs starting with correct one(125),
+         and check the confidence levels obtained from previous runs.
+         The 3rd param in assertAlmostEqual() has been optimised to 0.02.
+        """        
+
+        bpmEstimate = 125
+        expectedConfidence = 1
+        confidence = LoopBpmConfidence()(audio, bpmEstimate)        
+        self.assertAlmostEqual(expectedConfidence, confidence, 0.02)
```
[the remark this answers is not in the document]
cvf-bcn-gituser: The tolerance was actually 0.02. I had to increase it to 0.05 for the last test on line 115.
In any case, taking into  account that the length of the segment I chose , was way to short, I need to reconfigiure  parameter values for the ASSERT tests, as per comments below. Then I can investigate lowest consistent value for tolerance.

test/src/unittests/rhythm/test_loopbpmestimator.py
cvf-bcn-gituser: @ffont, while you are reading this comment, I wanted to comment something on a draft test for percivalbpmestimator

I have it in my local branch ( not yet in a PR).
In this, the audio subsections have been increased to take more significant parts of the loop for estimation.

https://github.com/cvf-bcn-gituser/essentia/blob/unittest_updates/test/src/unittests/rhythm/test_percivalbpmestimator.py

The estimation for full loop is  125.28408813476562, but the estimation for 3 different subsections (50% , 90%) is identical
at 124.90558624267578. This is different from loopbpmestiamtor which returns zero for the 90% section of the same audio.

src/algorithms/rhythm/loopbpmconfidence.cpp
```diff
@@ -54,66 +54,71 @@ void LoopBpmConfidence::compute() {
...
+    // Get original duration.
     int duration_samples = signal.size();
 
-    // Compute envelope
-    vector<Real> envelope;
-    _envelope->input("signal").set(signal);
-    _envelope->output("signal").set(envelope);
-    _envelope->compute();
+    // Check first that the signal is non-empty
+    if (duration_samples!=0){
```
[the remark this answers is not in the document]
cvf-bcn-gituser: Yes, Envelope has tests and they all pass.

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,157 @@
...
+         
+         Subsection [1000:6000], (Randomly chosen points, section 1)
+         BPM Estimate: 125: Confidence: 0.5275888442993164
+         BPM Estimate: 120: Confidence: 0.5464853048324585
+         BPM Estimate: 70: Confidence: 0.7354497313499451
+         Subsection [22000:38000], (Randomly chosen points, section 2)
+         BPM Estimate: 125: Confidence: 0.5117157697677612
+         BPM Estimate: 120: Confidence: 0.4512471556663513
+         BPM Estimate: 70: Confidence: 0.15343916416168213
+         The ASSERTs will check for these to two places of decimal
```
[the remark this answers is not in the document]
cvf-bcn-gituser: 90% of the loop results in the following estimates:
Subsection [0:len90]
         Estimate on the correct BPM:
         BPM Estimate: 125: Confidence: 0.19992440938949585
         Estimates on the incorrect BPMS:
         BPM Estimate: 120: Confidence: 0.4305669069290161
         BPM Estimate: 70: Confidence: 0.5011640191078186

Cutting1 % at beginning and 90% of the loop at the end results in the following estimates:         
Subsection [len01:len90],
         BPM Estimate: 125: Confidence: 0.9199735522270203
         BPM Estimate: 120: Confidence: 0.3631746172904968
         BPM Estimate: 70: Confidence: 0.7951852083206177

        I think the first test  (just 90%) is the best one to consider.
     
   Strangely , a high confidence results when we trim a bit at the beginning as well as the end.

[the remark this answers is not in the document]
ffont: Hi, IIRC the algorithm considers 3 cases in which a loop is not cut correctly but we still want to provide high confidence if the tempo estimation is good. This is when i) there is *non-musical** silence before the loop starts, ii) thee is non-musical silence after the loop ends, iii) there is non-musical silence at the start and AND the end of the loop. Because the algorithm matches the duration of the sound with the estimate of the BPM, non-musical silence at these points would invalidate the confidence value. What the algorithm does is to estimate the effective start and end points of the loop and also use these "alternative" durations in the comparison with the estimate of the BPM. I think the highest confidence is returned.

The algorithm does not account for cases in which the loop is cut in the middle (although if the loop is cut at a multiple of the estimated BPM beat duration, then the confidence should be high as well). I hope this clarifies.


* by non-musical I mean that silence is not part of the loop itself.

test/src/unittests/rhythm/test_loopbpmestimator.py
ffont: @dbogdanov I think it will be better if it always outputs a value, like `RhythmExtractor2013`. However I'm happy to update the extractor code if needed because it has not yet been deployed to Freesound. We're now working on new analysis system and I'd like to do a full re-analysis after that, using an updated extractor so now it is a good time to make changes if any are needed.

@cvf-bcn-gituser: this might be related to one of my comments above. IIRC loopbpmestiamtor combines the percival bpm estimator with the loop confidence, and outputs 0 if the confidence is below a threshold. The confidence estimator is not prepared for receiving 90% of the loop, it will return 0 (low confidence) if the length is away from a multiple of the estimated BPM beat duration. Does this explain the behaviour you're seeing?

cvf-bcn-gituser: @dbogdanov , @ffont is the final consensus now to return a value on a legit audio input, including silence or constant input make it consistent with RhythmExtractor2013?

test/src/unittests/rhythm/test_loopbpmconfidence.py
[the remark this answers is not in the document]
dbogdanov: @ffont Thanks! The documentation for LoopBpmConfidence could be improved to mention that we consider start/end silences. For now it only says "The confidence returned is based on comparing the duration of the signal with multiples of the BPM estimate." This can be misleading.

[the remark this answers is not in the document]
dbogdanov: @cvf-bcn-gituser @ffont We should be fine with using a single 90% segment for the regression test in LoopBpmEstimator, but should capture the 3 cases that can happen in test for LoopBpmConfidence. We can simulate them by adding artificial silence in the beginning or the end of audio loop. As I understand, given a BPM value, a certain amount of added silence should not change the output confidence unless it makes the new total audio length a better fit for the BPM value.

Does this imply that given a BPM value, adding artificial silence in the beginning or the end of the audio loop should not affect the resulting confidence value in LoopBpmConfidence?

test/src/unittests/rhythm/test_loopbpmestimator.py
dbogdanov: @cvf-bcn-gituser Yes, we all converge towards outputting 0 BPM in such cases, similar to RhythmExtractor2013.

test/src/unittests/rhythm/test_loopbpmconfidence.py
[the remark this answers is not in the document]
ffont: > @ffont Thanks! The documentation for LoopBpmConfidence could be improved to mention that we consider start/end silences. For now it only says "The confidence returned is based on comparing the duration of the signal with multiples of the BPM estimate." This can be misleading.

Ok, we can add the sentence: "The confidence returned is based on comparing the duration of the signal with multiples of the BPM estimate. The algorithm considers edge cases in which the signal includes non-musical silence at the start or end of the file (or both) and returns the confidence for the best fit." 
However we should refer to the paper for details, it is hard to give more details in the docstring. This is the link to the paper in case it is not there: https://repositori.upf.edu/handle/10230/33113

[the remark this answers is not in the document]
ffont: > We can simulate them by adding artificial silence in the beginning or the end of audio loop. As I understand, given a BPM value, a certain amount of added silence should not change the output confidence unless it makes the new total audio length a better fit for the BPM value.

Yes, I think this is how it works. The test would be that adding artificial silence at the start of the signal (or at the end or at both places) gives the same result as not adding any artificial silence.

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -128,25 +131,27 @@ def testEmpty(self):
...
-        onesAudio = ones(100)
+        audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))() 
+        onesAudio = ones(len(audio))
         bpmEstimate = 0
         confidence = LoopBpmConfidence()(onesAudio, bpmEstimate)
         self.assertEquals(0, confidence)
 
         # Non-zero estimate check results in non-zero confidence
-        # FIX-ME: Maybe given contant input, we should expect zero confidence
+        # FIXME: Maybe given contant input, we should expect zero confidence
```
dbogdanov, line 154: as @ffont mentioned, this happens because the estimation is based on audio length. Therefore, different constant input length will result in different estimations and it is part of normal behavior. We can append this explanation to the comments here and remove the FIXME.

dbogdanov (changes requested):

test/src/unittests/rhythm/test_loopbpmconfidence.py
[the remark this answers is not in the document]
dbogdanov: This paper is mentioned in the DOC. The new phrase looks very good, the only thing I would change is "file2 -> "audio signal"

test/src/unittests/rhythm/test_loopbpmestimator.py
cvf-bcn-gituser: Getting loopbpmestimator to converge with RhythmExtractor2013 will require updates in percivalbpmestimator.cpp and this is planned in separate PR.

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,212 @@
...
+        audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))() 
+        bpmEstimate = 125
+        confidence = LoopBpmConfidence()(audio, bpmEstimate)              
+        lenSilence = int(0.01*len(audio)) # Choose a silent length 1% of audio
+        silentAudio = zeros(lenSilence)
+
+        # case 1: there is non-musical* silence before the loop starts
+        signal1  = numpy.append(silentAudio, audio)
+        confidence = LoopBpmConfidence()(signal1, bpmEstimate)              
+        self.assertGreater(confidence, 0.9)
```
dbogdanov, line 146: I am trying to understand this test. What are the 0.9 and 0.27 values and what is reasoning for using `assertGreater`?

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,212 @@
...
+        # case 2: there is non-musical silence after the loop ends        
+        signal2  = numpy.append(audio,silentAudio)
+        confidence = LoopBpmConfidence()(signal2, bpmEstimate)  
+        self.assertGreater(confidence, 0.27) 
+
+
+        # case 3: there is non-musical silence at both ends
+        signal3 = numpy.append(signal1, silentAudio)
+        confidence = LoopBpmConfidence()(signal3, bpmEstimate)              
+        self.assertGreater(confidence, 0.27)
```
dbogdanov, line 164: As we have commented before, there are situations when adding a silence segment does not change the estimated confidence. This isn't covered by the test.

test/src/unittests/rhythm/test_loopbpmestimator.py
```diff
@@ -0,0 +1,113 @@
...
+
+class TestLoopBpmEstimator(TestCase):
+
+    def testInvalidParam(self):
+        self.assertConfigureFails(LoopBpmEstimator(), {'confidenceThreshold': -1})
+
+    def testRegression(self):
+        audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))()        
+      
+        expectedEstimate = 125
```
dbogdanov, line 33: What is the source of this ground truth for `techno_loop.wav` ? Is that something you manually checked?

dbogdanov (changes requested):

test/src/unittests/rhythm/test_loopbpmestimator.py
cvf-bcn-gituser: GT has been updated to 120 after a manual measurement using a stop watch,

test/src/unittests/rhythm/test_loopbpmconfidence.py
cvf-bcn-gituser: with the updates made to the groud truth , the addition of silence in the scenarios (beginning, end , both) now yields higher confidence.

test/src/unittests/rhythm/test_loopbpmconfidence.py
cvf-bcn-gituser: values have been updated with new GT.
The reasoning on assertGreater is to assure measured confidence is  greater than a benchmark that is reasonably high (0.8 to 0.9)

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,213 @@
...
+    https://freesound.org/people/Flowerpunkchip/sounds/458538/
+
+    Front end slience Confidence is  0.9915646314620972
+    Back end silence Confidence  is  0.9915646314620972
+    Silence Both end confidence is  0.9915646314620972    
+    """
+    def testSilentEdge(self):
+        audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))() 
+        bpmEstimate = 125
+        confidence = LoopBpmConfidence()(audio, bpmEstimate)
```
dbogdanov, line 138: Redundant line. `confidence` is overwritten at line 147.

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # Non-zero estimate check results in non-zero confidence
+        # The estimation is based on audio length.
+        # Different constant input length will result in different estimations.
+        bpmEstimate = 125
+        confidence = LoopBpmConfidence()(zeroAudio, bpmEstimate)
+        self.assertNotEquals(0, confidence)
+         
+    def testConstantInput(self):
+        audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))()
```
dbogdanov, line 197: Redundant line. This MonoLoader output is never used.

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,213 @@
...
+        # Non-zero estimate check results in zero confidence
+        # The estimation is based on audio length.
+        # Different constant input length will result in different estimations.
+        emptyAudio = []
+        bpmEstimate = 125
+        confidence = LoopBpmConfidence()(emptyAudio, bpmEstimate)                  
+        self.assertEquals(0, confidence)
+
+    def testZero(self):
+        audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))()
```
dbogdanov, line 183: Redundant line. This MonoLoader output is never used.

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,213 @@
...
+        confidence = LoopBpmConfidence()(signal1, bpmEstimate)      
+        self.assertGreater(confidence, benchmarkConfidence)
+
+    def testExactAudioLengthMatch(self):
+        audio = MonoLoader(filename=join(testdata.audio_dir, 'recorded', 'techno_loop.wav'))() 
+        bpmEstimate = 125
+        beatPeriod =  21168 # N.B The beat period is 21168 samples for 125 bpm @ 44.1k samp. rate
+        silentAudio = zeros(beatPeriod)
+        # Add non-musical silence to the beginnig of the audio 
+        signal1  = numpy.append(silentAudio, audio)
```
dbogdanov, line 156: Missing assert for `signal1`, it is never tested.

test/src/unittests/rhythm/test_loopbpmconfidence.py
```diff
@@ -0,0 +1,213 @@
...
+        benchmarkConfidence = 0.96 # This figure was arrived at emperically from the min. confidence observed with test runs 
+
+        # Test addition of non-musical silence before the loop starts
+        # The length is not a beat period,
+        # Nonetheless, we can stillreliably estimate the starting point because it is a hard transient.
+        signal1  = numpy.append(silentAudio, audio)
+        confidence = LoopBpmConfidence()(signal1, bpmEstimate)      
+        self.assertGreater(confidence, benchmarkConfidence)
+
+    def testExactAudioLengthMatch(self):
```
dbogdanov, line 150: I would add a test on zeros(beatPeriod * 4) which should also give confidence value of 1.

dbogdanov (changes requested):

cvf-bcn-gituser:
Is it possible to close request now?

dbogdanov:
This looks good by now! Thanks.

Issue #1108, opened 2021-02-18
REQ 16kHz support in EqualLoudness


jpauwels:
Feature request: it would be nice to be able to use EqualLoudness (and consequently EqLoudnessLoader) with signals sampled at 16kHz. Right now there's a big gap between the lowest and second lowest supported sample rates: 8000, 32000, 44100, 48000.

dbogdanov:
I agree this should be very useful as 16KHz is frequently used in many analysis applications. The filter values are hardcoded: https://github.com/MTG/essentia/blob/master/src/algorithms/filters/equalloudness.cpp#L54.

A pull request is welcome.

jpauwels:
I'd happily add it myself if I knew how those hardcoded values were calculated for the other samplerates.

dbogdanov:
@jpauwels We had a similar update in https://github.com/MTG/essentia/pull/362. There is a matlab script that can be used to compute the coefficients.

dbogdanov:
Thanks for the PR @jpauwels.  Can you send a CLA following our [contribution policy](http://essentia.upf.edu/contribute)?

jpauwels:
Done. FYI, the link you sent is dead, but I found a link to the form in the contributor guidelines

dbogdanov:
That was a bad copy-paste. Thanks!

PR #1110, opened 2021-02-19
Add support for a 16kHz sample rate to equal loudness filtering

jpauwels:
See #1108. I also fixed some warnings in the Matlab script and noticed that when I ran it with R2017b, the output is slightly different in its least significant digit than the stored values. I updated the `equal_loud_coef.txt` file to harmonise the documentation, but didn't update the actual values in the source code since this obviously will have zero effect in practice.


dbogdanov:
Thank you for contributing!

PR #1113, opened 2021-02-24
add small github actions script for a windows build

Bomme:
This is a short working example to play around with Github actions and the Windows runner.
Currently it only builds Eigen lib and successfully configures the build.

I put the Eigen lib in `packaging\win32_3rdparty` but it seems that this dir was used for the cross-built dependencies. Should we introduce a new dir, e.g. win64?

I need to explicitly enable [UTF-8 mode](https://docs.python.org/3/using/windows.html#utf-8-mode) for python because this special character `→` could not be printed. Maybe there is a nicer solution for this?

### DIFF
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -1,0 +1,41 @@
+name: Native Windows CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


Bomme:
@dbogdanov you can check how the actions build looks like here: https://github.com/Bomme/essentia/actions

I think the actions don't get triggered in this repository until the yml file is on a branch that is specifially mentioned, i.e. either we merge or we put this branch in this list if you want to have Actions enabled:
```
on:
  push:
    branches: [ master ]
```

PR #1117, opened 2021-03-08
Add jupyter notebooks with real-time examples

palonso:
- A real-time auto-tagging example
- A real-time system with four classifiers operating simultaneously


dbogdanov:
Merging this, but we should have a look at #1123 to improve tutorials/examples organization in the future.

PR #1118, opened 2021-03-08
Adding test for harmonic bpm

cvf-bcn-gituser:
Created 5 standard test cases for the Harmonic Bpm algorithm.
The test case "testZeros" needs further work, so FIXME is annotated.
Currently the algorithm hangs when an array of zeros are passed, putting PR in draft mode for now.


cvf-bcn-gituser:
On testing with more varied values  I noticed the program hands in the following scenario
        testBpms = [120, 120, 120, 120, 120, 120, 120,
                             240, 240, 240,240,240, 240, 240,
                             180, 180, 180, 180, 180, 180, 180,                        
                             90, 90, 90, 90, 90, 90, 90]
        harmonicBpms = HarmonicBpm(bpm=120,tolerance=0)(testBpms)   
There is a check on invalid configiration        
 self.assertConfigureFails(HarmonicBpm(), {'threshold': 0 })
 
 but not on invalid computation.

test/src/unittests/rhythm/test_harmonicbpm.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Perform a series of 4 RTs that vary parameters
+    def testRegressionStandard(self):
+        testBpms = [100, 101, 102, 103, 104]
+        harmonicBpms = HarmonicBpm(bpm=100)(testBpms)
+        expectedBpm = 100
+        self.assertEqual(harmonicBpms, expectedBpm)
+
+    # Check the value below threshold is not included in harmonicBpms
+    def testRegressionThresholdLimit(self):
```
dbogdanov, line 55: We should have a test where muliple BPMs are output, and some of those are discarded due to being below threshold.

test/src/unittests/rhythm/test_harmonicbpm.py
```diff
@@ -0,0 +1,91 @@
...
+        harmonicBpms = HarmonicBpm(bpm=100, threshold=99)(testBpms)
+        expectedBpm = 100
+        self.assertEqual(harmonicBpms, expectedBpm)
+
+    # Reduce tolerance parameter
+    def testRegressionLowTolerance(self):
+        testBpms = [100, 101, 102, 103, 104]
+        harmonicBpms = HarmonicBpm(bpm=100, tolerance=1)(testBpms)
+        expectedBpm = [100., 101.]
+        self.assertEqualVector(harmonicBpms, expectedBpm)
```
dbogdanov, line 66: Some inconsistency here: The default tolerance value in ` testRegressionStandard` was larger but only one BPM value was output in that test. Here we have smaller tolerance and multiple BPM outputs.

dbogdanov (changes requested):

cvf-bcn-gituser:
I just discovered a zero tolerance causes the program to hang, so I have new updates for the CPP to deal with this.
We may need to update the Esssentia documentation to say tolerance = 0 is not allowed.

test/src/unittests/rhythm/test_harmonicbpm.py
```diff
@@ -0,0 +1,91 @@
...
+
+class TestHarmonicBpm(TestCase):
+
+    def testInvalidParam(self):
+        # Test that we must give valid frequency ranges or order
+        self.assertConfigureFails(HarmonicBpm(), {'bpm': 0})
+        self.assertConfigureFails(HarmonicBpm(), {'threshold': 0 })
+        self.assertConfigureFails(HarmonicBpm(), {'tolerance': -1 })
+
+    def testConstantInput(self):
```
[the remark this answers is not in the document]
cvf-bcn-gituser: Its now called testEqualCandidateValues

test/src/unittests/rhythm/test_harmonicbpm.py
```diff
@@ -0,0 +1,91 @@
...
+        # Ensure that an exception is thrown if any element contains a zero
+        testBpms = [0, 100]
+        self.assertRaises(EssentiaException, lambda: HarmonicBpm()(testBpms))
+        testBpms = [100, 100, 100, 100, 0]
+        self.assertRaises(EssentiaException, lambda: HarmonicBpm()(testBpms))
+        testBpms = zeros(100)
+        self.assertRaises(EssentiaException, lambda: HarmonicBpm()(testBpms))
+
+    # Perform a series of 4 RTs that vary parameters
+    def testRegressionStandard(self):
```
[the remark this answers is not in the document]
cvf-bcn-gituser: I have added a new set of Test Cases

src/algorithms/rhythm/harmonicbpm.cpp
```diff
@@ -39,14 +39,26 @@ void HarmonicBpm::configure() {
   _threshold = parameter("threshold").toReal();
   _bpm = parameter("bpm").toReal();
   _tolerance = parameter("tolerance").toReal();
+  // This patch was added to fix a crash that happens when a zero tolerance value is configured
+  if (_tolerance==0){ 
+    throw(EssentiaException("HarmonicBpm: Zero tolerance found in Configuration"));
+  }
+  if (_bpm==0){ 
+    throw(EssentiaException("HarmonicBpm: Zero bpm found in Configuration"));
+  }
```
dbogdanov, line 48: From the user point of view I disagree with having these exceptions.

1) The algorithm should return all BPMs harmonically related to a given bpm. BPM=0 can be simply ignored in the list of the input BPMs instead of having an exception. What is the specific issue of having it in the input?  We can log a `E_INFO` message with a warning of having encountered a value close to 0 BPM.

    Btw, our `bpm` parameter is only allowed to be >=1. Another option, for consistency, would be to have that type of check for what input values to ignore. Also because we don't want to mess with very small BPM values that hardly have any musical sense.

2) Tolerance 0 means an ideal harmonic match. This might be a valid parameter value we can expect to work. For example, for BPM=120, the input value 60 is an ideal match. The algorithm shouldn't hang in the first place. Can you pinpoint the specific place and reason why it happens? 

    I see that `areEqual` method will always return False on 0 tolerance: `std::fabs(error)<tolerance`: https://github.com/MTG/essentia/blob/master/src/essentia/utils/bpmutil.h#L72. This `fabs(...)  < 0` will never happen. We could compare instead with `fabs(...) <= max(tolerance, numeric_limits<Real>::epsilon())`.


By the way, I see that our `bpm` parameter is integer, but there is not specific reason for that. We can change it to `Real` for a more generic use-case.

dbogdanov (changes requested):

src/algorithms/rhythm/harmonicbpm.cpp
cvf-bcn-gituser: regarding the hanging of the program in when tolerance = 0, it occurs in  HarmonicBpm::findHarmonicBpms(...)
There are 2 while loops and the tolerance =0  results in the i++ never being called.
A "watchdog" solution could be added to avoid the while loop hanging, but that would only treat the syntoms,
Best solution is a new areEqual method.
If bpm (integer) is to be changed to  bpm (real), I prefer do that in separate commit. This would also require documentation update (https://essentia.upf.edu/reference/std_HarmonicBpm.html).

dbogdanov: Yes, sure
> If bpm (integer) is to be changed to bpm (real), I prefer do that in separate commit. This would also require documentation update (https://essentia.upf.edu/reference/std_HarmonicBpm.html).

In fact, I just checked again and this is only a cosmetic change for documentation, because the integer value is read into Real inside the configure method: `_bpm = parameter("bpm").toReal();`. Just change the default value `60` to `60.0` and the documentation should render the `Real` type for this parameter.

src/algorithms/rhythm/harmonicbpm.cpp
```diff
@@ -55,12 +48,20 @@ vector<Real> HarmonicBpm::findHarmonicBpms(const vector<Real>& bpms) {
...
   if (_tolerance == 0){ 
     throw(EssentiaException("HarmonicBpm: Zero tolerance found"));
   }
 
+  // finding the minimum bpm and check it is above 20
+  int min = *min_element(bpms.begin(), bpms.end());
+  if (min < 20 ){
+    E_INFO ("HarmonicBpm: There are very low values among bpms candidates (below 20 BPM)");
+  }
+
```
dbogdanov, line 64: I am not sure about this threshold (=20). It is reasonable that a BPM of 20 is weirdly slow, but it is a sub-harmonic of common BPMs that can occur (40, 60, 80, etc.) and in theory it can occur in some BPM estimation algorithms. For the `bpm` parameter we require >= 1. Why not using the same?

dbogdanov (changes requested):

cvf-bcn-gituser:
I merged from trunk, added in my local changes on top.
Did initial Unit Test.
Still need fine tuning and testing on the testZeros.
Prefer to continue on Linux machine

src/algorithms/rhythm/harmonicbpm.cpp
cvf-bcn-gituser: I split the warnings statement scenarios into:
(1) low BPM (to cover the possible sub harmonic range below 20 when threshold is also set very low
(2) extremely low BPM for value equal or below 1.

cvf-bcn-gituser:
some of the unit test outputs have changed with new CPP. Putting in draft mode for now.

cvf-bcn-gituser:
checked that calculated values are OK in testRegressionTwoOctaveRange, testRegressionMultipleValues are consistent with the expected calculations in the updated harmonicbpm.cpp.

src/algorithms/rhythm/harmonicbpm.cpp
```diff
@@ -44,6 +44,21 @@ void HarmonicBpm::configure() {
 
 
 vector<Real> HarmonicBpm::findHarmonicBpms(const vector<Real>& bpms) {
+  // A zero element in the bpms vector will cause program to hang.
+  // Ensure this value doesnt exist.
+  if (std::count(bpms.begin(), bpms.end(), 0)){
```
cvf-bcn-gituser, line 49: Modify the check that it is not below or equal to zero.

src/algorithms/rhythm/harmonicbpm.cpp
```diff
@@ -44,6 +44,21 @@ void HarmonicBpm::configure() {
...
+     throw(EssentiaException("HarmonicBpm: Zero bpm value found"));
+  }
+
+  // finding the minimum bpm and check it is above thresholds
```
cvf-bcn-gituser, line 53: discard all values below 1 and add an E_INFO

src/algorithms/rhythm/harmonicbpm.cpp
```diff
@@ -44,6 +44,21 @@ void HarmonicBpm::configure() {
...
+  Real min = *min_element(bpms.begin(), bpms.end());
+  if ((min > 1) && (min < 20)){
+     E_INFO ("HarmonicBpm: There are very low values among bpms candidates (below 20 BPM)");
+  } 
+  else if (min < 1){
+     E_INFO ("HarmonicBpm: There are extremely low values among bpms candidates (equal or below 1 BPM)");
+  }
```
cvf-bcn-gituser, line 60: This can be removed

src/algorithms/rhythm/harmonicbpm.cpp
```diff
@@ -52,12 +67,12 @@ vector<Real> HarmonicBpm::findHarmonicBpms(const vector<Real>& bpms) {
 
   // Discard BPM values with GCD below the threshold.
   for (int i=0; i<int(bpms.size()); i++) {
-    Real gcd = greatestCommonDivisor(_bpm, bpms[i], _tolerance);
-    if (gcd >= _threshold) {
-      cout << gcd << bpms[i] << endl;
-      harmonicBpms.push_back(bpms[i]);
-      if (gcd < mingcd) mingcd = gcd;
-    }
+     Real gcd = greatestCommonDivisor(_bpm, bpms[i], _tolerance);
```
cvf-bcn-gituser, line 70: //cout

test/src/unittests/rhythm/test_harmonicbpm.py
```diff
@@ -0,0 +1,190 @@
...
+    # Check that illegal parameters in configuration raise asserts
+    def testInvalidParam(self): 
+        # Test that we must give valid frequency ranges, thresholds and tolerance.
+        self.assertConfigureFails(HarmonicBpm(), {'bpm': -1})
+        self.assertConfigureFails(HarmonicBpm(), {'bpm': 0})
+        self.assertConfigureFails(HarmonicBpm(), {'threshold': -1 })
+        self.assertConfigureFails(HarmonicBpm(), {'threshold': 0 })
+        self.assertConfigureFails(HarmonicBpm(), {'tolerance': -1 })
+
+    # Check that illegal parameters in computation raise asserts.
```
cvf-bcn-gituser, line 40: remove

test/src/unittests/rhythm/test_harmonicbpm.py
```diff
@@ -0,0 +1,190 @@
...
+
+    # Check multiple BPMs with diverse octaves ranges
+    def testRegressionMultipleValues(self):
+        testBpms = [120, 120, 120, 120, 120, 120, 120,
+                             240, 240, 240, 240, 240, 240, 240,
+                             180, 180, 180, 180, 180, 180, 180,  
+                             90, 90, 90, 90, 90, 90, 90]
+        
+        # Check with default threshold (20) and BPM = 120
+        expectedHarmonicBps = [90, 120, 180, 240]
```
cvf-bcn-gituser, line 140: expectedHarmonicBps  rename to expectedHarmonicBpms

test/src/unittests/rhythm/test_harmonicbpm.py
```diff
@@ -0,0 +1,190 @@
...
+        #  We want to ensure nothing bad happens when tolerance = 0.
+        harmonicBpms = HarmonicBpm(bpm=60, tolerance=0, threshold=20)(testBpms)
+        self.assertEqualVector(harmonicBpms, expectedHarmonicBps)
+        #  Check range tolerance values of multiples of 5.
+        harmonicBpms = HarmonicBpm(bpm=60, tolerance=5, threshold=20)(testBpms)
+        self.assertEqualVector(harmonicBpms, expectedHarmonicBps)
+        harmonicBpms = HarmonicBpm(bpm=60, tolerance=10, threshold=20)(testBpms)
+        self.assertEqualVector(harmonicBpms, expectedHarmonicBps)
+        harmonicBpms = HarmonicBpm(bpm=60, tolerance=20, threshold=20)(testBpms) 
+        self.assertEqualVector(harmonicBpms, expectedHarmonicBps)
```
cvf-bcn-gituser, line 176: add last test case so that 10 is included, with threshold = 1 (all in case)

test/src/unittests/rhythm/test_harmonicbpm.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def testZeros(self):
```
cvf-bcn-gituser, line 178: rename to mention invalid bpms, (below 1 values)

PR #1121, opened 2021-03-11
Introduce new unit test files for pitch salience

cvf-bcn-gituser:
Adding new scripts for testing pitch salience.
They contain the regular spot checks testInvalidParam, testEmpty, etc and one " testARealCase" with default params.
This initial draft is based on using as real input the audio file "vignesh.wav". This may be supplemented/replaced with other audio test files or with synthetic files to get more test coverage. The "testARealCase" is really a regression test using previous snapshot measurements as a baseline for detecting if future changes modify algorithm behaviour.


cvf-bcn-gituser:
@dbogdanov, almost ready for review. Early feedback required to help know what to do with subsequent variants of  PitchContours (PitchContourMonoMelody, etc)
Pre-checks can start.

cvf-bcn-gituser:
In draft status, but comments welcome.

cvf-bcn-gituser:
TODO: update copyright to 2021

cvf-bcn-gituser:
@dbogdanov , some small updates on  Unit Tests: Removed resets. Increased precision to 8 (9cccfc9 )

https://github.com/MTG/essentia/pull/1121/commits/9cccfc944933122fbeda1a323094147d17906c4f

cvf-bcn-gituser:
In commit  "32f1a77a04d0c3b147892af5de8ce046bfa2d4f7" I added the artificial music generator and signal generator for testing.
There is a TODO in the part where I added the music generator, since the output was not what was expected.
The easiest fix will probably be to replace all the functions with a simpler scale or ladder generator used else where in example.
(Technically I should revert this to draft, but I wont do that since I expect 1 set of  follow up comments)

cvf-bcn-gituser:
@dbogdanov Music Generator replaced with raw scale ladder. Only 1 TODO remaining concerning doubts on low confidence output.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,79 @@
...
+    def testZero(self):
+        freqs = zeros(1024)
+        mags  = zeros(1024)
+        self.assertRaises(RuntimeError, lambda: PitchSalienceFunction()(freqs, mags))
+
+    def testEmpty(self): 
+        self.assertEqualVector(PitchSalienceFunction()([], []), zeros(600))
+
+    def testOne(self):        
+        self.assertEqualVector(PitchSalienceFunction()(ones(1024), ones(1024)), zeros(600))
```
dbogdanov, line 46: This test also doesn't make much sense (it gives 1024 peaks at 1 Hz below the minimum frequency of 55 Hz, with the amplitude 1). 

Better tests would be:
- Provide a single input peak with a unit magnitude at the reference frequency, and validate that the output salience function has only one non-zero element at the first bin.
- Provide multiple duplicate peaks at the reference frequency. 
- Provide a single input peak below the reference frequency, so that the result is an empty pitch salience function
- How does a single input peak above the maximum bin affect the salience function? You would see the activations at the subharmonics of the input frequency, that fall within the 6000-cent frequency range.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def testUnequalInputs(self):
+        # Choose a sample set of frequencies and magnitude vectors of unqual length
+        freqs = [250, 500, 1000, 2000, 3500] # length 5
+        mags = [0.5, 0.5, 0.5, 0.5] # length 4
+        self.assertRaises(EssentiaException, lambda: PitchSalienceFunction()(freqs, mags))
+
+    def testARealCase(self):
```
dbogdanov, line 54: >  The salience of a given frequency is computed as the sum of the weighted energies found at integer multiples (harmonics) of that frequency.

We are missing a test checking this. For example,
- Give a single peak as an input and validate how all subharmonics are activated in the salience function.
- Create a simple test with only a few input peaks. 

The [weights of harmonics](https://github.com/MTG/essentia/blob/master/src/algorithms/tonal/pitchsaliencefunction.cpp#L57) are defined by parameter `harmonicWeight`. 
- We can test the default 0.8 value.
- Test a simpler value of 1 and 0.

The values 0 and 1 aren't currently allowed for `harmonicWeight`, but we can enable those  as they may be useful in some practical use cases. They are also useful for running tests. We can run simple tests proposed in this comment  for both 0 and 1 values, apart from the default 0.8 value. Furthermore, the documentation already mentioned the case of `=1 for no decay` even though is not enabled.

Note that to make `harmonicWeight = 0` work, we need to adjust the code here: https://github.com/MTG/essentia/blob/master/src/algorithms/tonal/pitchsaliencefunction.cpp#L57
This is because the value of pow(0, 0) has undefined behavior. We should manually set the pow(0,0) value to 1. We can do it after the loop, setting `_harmonicWeights[0] = 1` in the case if `_harmonicWeight` == 0.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,79 @@
...
+        sr = 44100
+        hopSize = 512
+        filename = join(testdata.audio_dir, 'recorded', 'vignesh.wav')
+        audio = MonoLoader(filename=filename, sampleRate=44100)()
+        # Get the frequencies and magnitudes of the spectral peaks
+        freq_speaks, mag_speaks = SpectralPeaks()(audio)
+        # Start with default params
+        psf = PitchSalienceFunction()
+        calculatedPitchSalience = psf(freq_speaks,mag_speaks)
+
```
dbogdanov, line 65: This is a wrong way of using SpectralPeaks. It is computed on spectrum on a frame level. Have a look at the example here: https://github.com/MTG/essentia/blob/master/src/examples/tutorial/example_predominantmelody_by_steps.py#L65

dbogdanov: PitchSalienceFunction is computed on separate frames. Thus, we can assess it on a few random frames, or the entire track, but only if the reference NPY array is not very large.

dbogdanov (changes requested):
Left feedback related to PitchSalienceFunction.

test/src/unittests/tonal/test_pitchsaliencefunction.py
cvf-bcn-gituser: I have addressed the other comments and will make a local commit on them.
This comment requires more work.

I will propose to close this PR and I split this PR into two parts:
1 part taking 3 algorithsm and other part taking 2 algorithms

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
+        self.assertEqualVector(startTimes, [])
+        # Previous tests showed small duration of 0.0058 seconds for zero or empty inputs.        
+        self.assertAlmostEqual(duration, 0.0058, 8)
+
+    def testZeros(self):
+        bins, saliences, startTimes, duration = PitchContours()(array(zeros([2,256])), array(zeros([2,256])))      
+        self.assertEqualVector(bins, [])
+        self.assertEqualVector(saliences, [])
+        self.assertEqualVector(startTimes, [])
+        # Previous tests showed small duration of 0.0058 seconds for zero or empty inputs.
```
dbogdanov, line 50: Same here, we can precompute the expected duration from the input number of frames.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
+        self.assertAlmostEqual(duration, 0.0058, 8)
+
+    def testZerosUnequalInputs(self):        
+        peakBins = [zeros(4096), zeros(4096)]
+        peakSaliences = [zeros(1024), zeros(1024)]
+        self.assertRaises(RuntimeError, lambda: PitchContours()(peakBins, peakSaliences))
```
dbogdanov, line 56: Both `testZerosUnequalInputs` and `testOnesUnequalInputs` are redundant and should be replaced with a single `testUnequalInputs` that tests the exception when input vectors differ in size.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def testOnes(self):
+        peakBins = [ones(4096),ones(4096)]
+        peakSaliences = [ones(4096),ones(4096)]
+        bins, saliences, startTimes, duration = PitchContours()(peakBins, peakSaliences)
+        self.assertEqualVector(bins, [])
+        self.assertEqualVector(saliences, [])
+        self.assertEqualVector(startTimes, [])
+        self.assertAlmostEqual(duration, 0.0058, 3)
```
dbogdanov, line 65: This test doesn't make much sense to me: the input is two frames of 4096 peaks at the first cent bin each. Perhaps what you want to test instead is a single pitch contour at a fixed frequency bin with a fixed salience of 1.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
+        
+    def testOnesUnequalInputs(self):
+        peakBins = [ones(4096), ones(4096)]
+        peakSaliences = [ones(1024), ones(1024)]
+        self.assertRaises(RuntimeError, lambda: PitchContours()(peakBins, peakSaliences))
```
dbogdanov, line 70: Redundant (commented above)

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
+
+        bins, saliences, startTimes, duration = pc(peakBins, peakSaliences)
+
+        #The reason is that 2D vectors dont reload easily from files, using the method shown below.
+        save('pitchcontourstarttimes_synthetic.npy',startTimes) 
+        save('pitchcontourbins_synthetic.npy', bins)               
+        save('pitchcontoursaliences_synthetic.npy', saliences)
+
+        # Captured from previous runs of pitch contours on the synthesis
+        expectedDuration = 0.00290249427780509
```
dbogdanov, line 121: Same here, we should deduce this duration from the input audio signal instead of relying on a previous run.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
+        spectrum = Spectrum()
+        spectralpeaks = SpectralPeaks()
+        pc = PitchContours()
+
+        # Populate an array of frame-wise vectors of cent bin values representing each contour
+        peakBins = []
+        # Populate a frame-wise array of values of salience function peaks
+        peakSaliences = []        
+        for frame in FrameGenerator(audio, frameSize=1024, hopSize=hopSize,
+                                    startFromZero=True):
```
dbogdanov, line 99: How long is the audio? it is just one frame (`frame1 + frame2`), so this FrameGenerator wouldn't do many iterations.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
+    def testOnesUnequalInputs(self):
+        peakBins = [ones(4096), ones(4096)]
+        peakSaliences = [ones(1024), ones(1024)]
+        self.assertRaises(RuntimeError, lambda: PitchContours()(peakBins, peakSaliences))
+
+    def testRegressionSynthetic(self):
+        # Use synthetic audio for Regression Test. This keeps NPY files size low.     
+        sr = 44100
+        frameSize = 1024
+        hopSize = 512
```
dbogdanov, line 76: How do you decide on these parameters? The recommended settings in the DOC are: 
> Recommended processing chain: (see [1]): EqualLoudness -> frame slicing with sample rate = 44100, frame size = 2048, hop size = 128 -> Windowing with Hann, x4 zero padding -> Spectrum -> SpectralPeaks -> PitchSalienceFunction (10 cents bin resolution) -> PitchSalienceFunctionPeaks.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
+           index+=1         
+
+        self.assertAlmostEqualVectorFixedPrecision(expectedPitchContourStartTimes,startTimes, 8)        
+        self.assertAlmostEqual(expectedDuration,duration, 8)             
+
+ 
+    def testARealCase(self):
+        frameSize = 1024
+        sr = 44100
+        hopSize = 512
```
dbogdanov, line 148: Same here. We could go for the default recommended settings instead.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
+        for frame in FrameGenerator(audio, frameSize=1024, hopSize=hopSize,
+                                    startFromZero=True):
+            freq_peaks, mag_peaks = spectralpeaks(spectrum(w(frame)))
+            len_freq_peaks = len(freq_peaks)
+            len_mag_peaks = len(mag_peaks)
+            
+            # Skip the first vector elements that correspond to zero frequency
+            freq_peaksp1 = freq_peaks[1:len_freq_peaks]
+            # Do the same for mag. vector to keep the lengths the same
+            mag_peaksp1 = mag_peaks[1:len_mag_peaks]
```
dbogdanov, line 173: We don't need to skip those, because we expect the PitchSalienceFunction to ignore peaks below its reference frequency.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,216 @@
...
```
dbogdanov, line 107: We don't need to skip those, because we expect the PitchSalienceFunction to ignore peaks below its reference frequency.

dbogdanov (changes requested):
We should further consider adding test for `pitchContinuity` and `timeContinuity`.

cvf-bcn-gituser:
replaced by #1142 and another new PR will be made for pitch contours

test/src/unittests/tonal/test_pitchcontours.py
cvf-bcn-gituser: FC has been removed

PR #1137, opened 2021-04-14
New contour (mono and multi) Test Cases

cvf-bcn-gituser:
New Test Cases based on the  test_pitchcontoursmelody.py, for multi-melody and mono-melody variants.
Putting in draft mode but comments welcome.

PR #1141, opened 2021-04-19
Test Cases for  MultiPitchKlapuri and MultiPitchMelodia

cvf-bcn-gituser:
Similar to PR 1137
https://github.com/MTG/essentia/pull/1137
Extending test coverage to consider multipitch algorithms.


cvf-bcn-gituser:
@dbogdanov , updated to add melodia. this algorithm has a very similar  i/o. It is the monophonic pre-decessor of Kalpuri.
Still in draft mode but comments welcome.

cvf-bcn-gituser:
There is a FIXME in the Unit Tests where a crash occurs.
The origin of the DIV by zero error is not in pitchmelodia.cpp. Its origin is "PSF" algorithm chain.
This is being dealt with in PR # 1142.

PR #1142, opened 2021-04-30
Rewritten tests for PitchSalienceFunction and  PitchSalienceFunctionPeaks

cvf-bcn-gituser:
The  regression test for a real audio (vignesh audio) in test_pitchsaliencefunction.py  also considers testing the PitchSalienceFunctionPeaks  algorithm in the loop and compares to a reference 
output from PredominantPitchMelodia.

There are use of reference  NPY files in this version


test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,223 @@
...
+    def testEmpty(self): 
+        self.assertEqualVector(PitchSalienceFunction()([], []), zeros(600))
+
+    # Provide a single input peak with a unit magnitude at the reference frequency, and 
+    # validate that the output salience function has only one non-zero element at the first bin.
+    def testSinglePeak(self):
+        freq_speaks = [55] # length 1
+        mag_speaks = [1] # length 4
+        calculatedPitchSalience = PitchSalienceFunction()(freq_speaks,mag_speaks)    
+        self.assertEqual(calculatedPitchSalience[0],1)
```
dbogdanov, line 46: This test does not ensure that the rest of the elements are zeros.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def test3Peaks(self):
+        freq_speaks = [55, 100, 340] # length 1
+        mag_speaks = [1, 1, 1] # length 4
+        # For a single frequency 55 Hz with unitary  amplitude the 10 non zero salience function values are the following
+        calculatedPitchSalience = PitchSalienceFunction()(freq_speaks,mag_speaks)    
+        self.assertAlmostEqual(calculatedPitchSalience[0], 1.16384, 3)
```
dbogdanov, line 53: What about the rest of the elements apart from the index 0? They should be covered by the test too.

cvf-bcn-gituser (commented):
remove comments on lines 51 and 59

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,223 @@
...
+        self.assertEqual(len(calculatedPitchSalience), 600)        
+
+    def test3PeaksHw0(self):
+        freq_speaks = [55, 100, 340] # length 1
+        mag_speaks = [1, 1, 1] # length 4
+        # For a single frequency 55 Hz with unitary  amplitude the 10 non zero salience function values are the following
+
+        calculatedPitchSalience = PitchSalienceFunction(harmonicWeight=0)(freq_speaks, mag_speaks)    
+        self.assertAlmostEqual(calculatedPitchSalience[0], 1.16384, 3)
```
dbogdanov, line 62: Just to clarify the algorithm: Here if `harmonicWeight = 0` should we expect the output value at the 0th bin to be 1?

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def testDifferentPeaks(self):
+        freq_speaks = [55, 85] # length 1
+        mag_speaks = [1, 1] # length 4
+        calculatedPitchSalience1 = PitchSalienceFunction()(freq_speaks,mag_speaks)
+              
+        mag_speaks = [0.5, 2] # length 4
+        calculatedPitchSalience2 = PitchSalienceFunction()(freq_speaks,mag_speaks)
+        self.assertAlmostEqual(calculatedPitchSalience1[0], 0.5, 3)        
+        self.assertAlmostEqual(calculatedPitchSalience2[0], 1.5, 3)
```
dbogdanov, line 80: Here the rest of the values in the output vector aren't covered by the tests as well.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def testBelowReferenceFrequency1(self):
+        freq_speaks = [50] # length 1
+        mag_speaks = [1] # length 4
+        expectedPitchSalience = zeros(600)
+        calculatedPitchSalience = PitchSalienceFunction()(freq_speaks,mag_speaks)        
+        self.assertEqualVector(calculatedPitchSalience, expectedPitchSalience)
+
+    # Provide multiple duplicate peaks at the reference frequency.
+    def testBelowReferenceFrequency2(self):
```
dbogdanov, line 90: The comment does not correspond to the test.

dbogdanov: The test mentioned in the comment is missing.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,223 @@
...
+            salience_peaks_saliences_array.append(salience_peaks_saliences)
+
+        bins, saliences, startTimes, duration = run_pitch_contours(salience_peaks_bins_array, salience_peaks_saliences_array)
+        pitch1, pitchConfidence1 = run_pitch_contours_melody(bins, saliences, startTimes, duration) 
+        pitch2, pitchConfidence2 = run_predom_pitch_melodia(audio)
+
+        self.assertAlmostEqualVectorFixedPrecision(pitch1[:600], pitch2[:600], 8)
+        #TODO why do the outputs align up to a certain array point?
+        self.assertAlmostEqualVectorFixedPrecision(pitch1[:721], pitch2[:721], 8)
+        #self.assertAlmostEqualVectorFixedPrecision(pitchConfidence1, pitchConfidence2, 8)
```
dbogdanov, line 163: This test is running PitchContour* algorithms, however this not what are expected to test here. We want to have a regression test validating that the values of the PitchSalienceFunction output are correct.

The type of test implemented here should go to the algorithms related to tracking pitch contours.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,223 @@
...
+            frame = run_windowing(frame)
+            spectrum = run_spectrum(frame)
+            peak_frequencies, peak_magnitudes = run_spectral_peaks(spectrum)
+            salience = run_pitch_salience_function(peak_frequencies, peak_magnitudes)
+            salience_peaks_bins, salience_peaks_saliences = run_pitch_salience_function_peaks(salience)
+                
+        expectedBins = [270., 150. ,80. ,30.]
+        expectedPeaks =  [0.09777679, 0.07822143, 0.06257715, 0.05006172]
+        self.assertAlmostEqualVectorFixedPrecision(expectedBins, salience_peaks_bins, 6)
+        self.assertAlmostEqualVectorFixedPrecision(expectedPeaks, salience_peaks_saliences, 6)
```
dbogdanov, line 218: Same here. We are testing PitchSalienceFunction, we can do it directly on its output. The tests for peaks are just an addition on top.

Perhaps we could have a npy with the expected matrix output if it isn't too large in filesize?

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,174 @@
...
+  
+    def testInvalidParam(self):
+        self.assertConfigureFails(PitchSalienceFunctionPeaks(), {'binResolution': -1})
+        self.assertConfigureFails(PitchSalienceFunctionPeaks(), {'maxFrequency': -1})
+        self.assertConfigureFails(PitchSalienceFunctionPeaks(), {'minFrequency': -1})
+        self.assertConfigureFails(PitchSalienceFunctionPeaks(), {'referenceFrequency': -1})                
+
+    def testEmpty(self): 
+        self.assertRaises(RuntimeError, lambda: PitchSalienceFunctionPeaks()([]))
+
```
dbogdanov, line 34: Here we are missing a test to validate that the input vector size corresponds to the `binResolution`, `referenceFrequency`, and `maxFrequency`. What would happen if the input size is [smaller that expected](https://github.com/MTG/essentia/blob/master/src/algorithms/tonal/pitchsaliencefunctionpeaks.cpp#L47) according to these parameters?

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,174 @@
...
+        self.assertConfigureFails(PitchSalienceFunctionPeaks(), {'maxFrequency': -1})
+        self.assertConfigureFails(PitchSalienceFunctionPeaks(), {'minFrequency': -1})
+        self.assertConfigureFails(PitchSalienceFunctionPeaks(), {'referenceFrequency': -1})                
+
+    def testEmpty(self): 
+        self.assertRaises(RuntimeError, lambda: PitchSalienceFunctionPeaks()([]))
+    
+    # Check for empty output arrays for a salience function with all zero values
+    def testZeros(self): 
+        bins, values = PitchSalienceFunctionPeaks()(zeros(128))
```
dbogdanov, line 37: We should use the default expected size of the input vector.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,174 @@
...
+        self.assertEqualVector(bins, [])
+        self.assertEqualVector(values, [])    
+
+    # Check for empty output arrays for a salience function repeated values of different lengths
+    def testConstantInputs(self): 
+        bins, values = PitchSalienceFunctionPeaks()([0.5, 0.5, 0.5])
```
dbogdanov, line 43: This is a pitch salience function with 3 frequency bins only. This doesn't make much sense.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,174 @@
...
+        self.assertConfigureFails(PitchSalienceFunctionPeaks(), {'referenceFrequency': -1})                
+
+    def testEmpty(self): 
+        self.assertRaises(RuntimeError, lambda: PitchSalienceFunctionPeaks()([]))
+    
+    # Check for empty output arrays for a salience function with all zero values
+    def testZeros(self): 
+        bins, values = PitchSalienceFunctionPeaks()(zeros(128))
+        self.assertEqualVector(bins, [])
+        self.assertEqualVector(values, [])
```
dbogdanov, line 39: PitchSalienceFuncionPeaks is just a wrapper for PeakDetection, so these tests should be already covered by PeakDetection itself (we can check) and are redundant here.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,174 @@
...
+    
+    # As above but with Harmonic Weight = 1
+    def testSinglePeakHw1(self):
+        freq_speaks = [55] # length 1
+        mag_speaks = [1] # length 4
+        calculatedPitchSalience = PitchSalienceFunction(harmonicWeight=1)(freq_speaks,mag_speaks)    
+        bins, values = PitchSalienceFunctionPeaks()(calculatedPitchSalience)
+        self.assertEqualVector(bins, [0.])
+        self.assertEqualVector(values, [1.])   
+        #self.assertEqual(calculatedPitchSalience[0],1)
```
dbogdanov, line 79: There's no harmonics, only a single frequency peak as an input, therefore testing both harmonicWeight = 0 and 1 is redundant.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Same as above but using a non-default config parameter.
+    def testBelowReferenceFrequency2(self):
+        freq_speaks = [30] 
+        mag_speaks = [1] 
+        expectedPitchSalience = zeros(600)
+        calculatedPitchSalience = PitchSalienceFunction(referenceFrequency=40)(freq_speaks,mag_speaks)        
+        bins, values = PitchSalienceFunctionPeaks()(calculatedPitchSalience)
+        self.assertEqualVector(bins, [])
+        self.assertEqualVector(values, [])
```
dbogdanov, line 169: Redundant? This has been covered in PitchSalienceFunction and PeakDetection should have a test for zero input.

dbogdanov (changes requested):
Reviewed PitchSalienceFunction and PitchSalienceFunctionPeaks.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,138 @@
...
+        
+    def testEmpty(self):
+        emptyPeakBins = []
+        emptyPeakSaliences = []
+        bins, saliences, startTimes, duration = PitchContours()(emptyPeakBins, emptyPeakSaliences)       
+        self.assertEqualVector(bins, [])
+        self.assertEqualVector(saliences, [])
+        self.assertEqualVector(startTimes, [])
+        # Previous tests showed small duration of 0.0058 seconds for zero or empty inputs.        
+        calculatedDuration = (2*defaultHopSize)/defaultSampleRate
```
dbogdanov, line 74: This is unclear. We need to investigate why it happens.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,138 @@
...
+        emptyPeakBins = [[],[]]
+        emptyPeakSaliences = [[],[]]
+        theHopSize= 2*defaultHopSize
+        bins, saliences, startTimes, duration = PitchContours(hopSize=theHopSize)(emptyPeakBins, emptyPeakSaliences)       
+        self.assertEqualVector(bins, [])
+        self.assertEqualVector(saliences, [])
+        self.assertEqualVector(startTimes, [])
+        # Previous tests showed small duration of 0.0058 seconds for zero or empty inputs.        
+        calculatedDuration = (2*theHopSize)/defaultSampleRate
+        self.assertAlmostEqual(duration, calculatedDuration, 8)
```
dbogdanov, line 87: This is a huge relative error threshold. Why is that?

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,138 @@
...
+        # Previous tests showed small duration of 0.0058 seconds for zero or empty inputs.
+        calculatedDuration = (2*theHopSize)/defaultSampleRate            
+        self.assertAlmostEqual(duration, calculatedDuration, 8)
+
+    #256 frames x single zero-salience peak at the same cent bin in each frame.
+    def testSingleZeroSaliencePeak(self):
+        peakBins = array(zeros([1, 256]))
+        peakSaliences = array(zeros([1, 256]))
+
+        theHopSize = 4*defaultHopSize
```
dbogdanov, line 104: It is unclear why we need to change the default hop size recommended for pitch contour estimation.

test/src/unittests/tonal/test_pitchsaliencefunction.py
cvf-bcn-gituser: ok I propose something like

 //the value of pow(0, 0) has undefined behavior. When weights is zero set the pow(0,0) value to 1
  if (_harmonicWeight==0){
      _harmonicWeights.push_back(1);
       for (int h=1; h<600; h++) {
          _harmonicWeights.push_back(0);
       }
  }

src/algorithms/tonal/pitchsaliencefunction.cpp
```diff
@@ -53,8 +53,15 @@ void PitchSalienceFunction::configure() {
 
   _harmonicWeights.clear();
   _harmonicWeights.reserve(_numberHarmonics);
-  for (int h=0; h<_numberHarmonics; h++) {
-    _harmonicWeights.push_back(pow(_harmonicWeight, h));
+
+  //the value of pow(0, 0) has undefined behavior. When weights is zero set the pow(0,0) value to 1
+  if (_harmonicWeight==0){
+      _harmonicWeights.push_back(1);
```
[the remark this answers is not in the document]
cvf-bcn-gituser: ok I propose something like

 //the value of pow(0, 0) has undefined behavior. When weights is zero set the pow(0,0) value to 1
  if (_harmonicWeight==0){
      _harmonicWeights.push_back(1);
       for (int h=1; h<600; h++) {
          _harmonicWeights.push_back(0);
       }
  }

test/src/unittests/tonal/test_pitchsaliencefunction.py
cvf-bcn-gituser: then this need npy files

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,138 @@
...
+
+        calculatedDuration = (2*theHopSize)/defaultSampleRate            
+        self.assertAlmostEqual(duration, calculatedDuration, 8)
+
+    def testUnequalInputs(self):        
+        peakBins = [zeros(4096), zeros(4096)]
+        peakSaliences = [zeros(1024), zeros(1024)]
+        self.assertRaises(RuntimeError, lambda: PitchContours()(peakBins, peakSaliences))
+    
+suite = allTests(TestPitchContours)
```
dbogdanov, line 134: The pitch ladder tests make sense for this particular algorithm instead of PitchSalienceFunctionPeaks.

dbogdanov (changes requested):
Reviewed PitchContours

test/src/unittests/tonal/test_pitchsaliencefunction.py
cvf-bcn-gituser: The philosophy was to create an end to end test, that would YES involve other algorithms, but would have the advantage of not needing NPY files.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,138 @@
...
+        
+        theHopSize= 4 * defaultHopSize        
+        _,  _, _, duration = PitchContours(hopSize=theHopSize)(emptyPeakBins, emptyPeakSaliences)               
+        calculatedDuration = (2 * theHopSize)/defaultSampleRate        
+        self.assertAlmostEqual(duration, calculatedDuration, 8)
+
+        theHopSize= 8 * defaultHopSize        
+        _,  _, _, duration = PitchContours(hopSize=theHopSize)(emptyPeakBins, emptyPeakSaliences)               
+        calculatedDuration = (2 * theHopSize)/defaultSampleRate        
+        self.assertAlmostEqual(duration, calculatedDuration, 8)
```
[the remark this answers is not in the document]
cvf-bcn-gituser: output duration is now zero for empty inputs. 
Bug was fixed in CPP.

test/src/unittests/tonal/test_pitchcontours.py
cvf-bcn-gituser: duration is calculated as follows in the code:

The test wants to examine different duration outcomes with different hop sizes
duration = _numberFrames * _frameDuration;

test/src/unittests/tonal/test_pitchcontours.py
cvf-bcn-gituser: testRegressionSynthetic is added.
The synthetic outputs of  contours_bins, contours_saliences, contours_start_times are saved in NPY files

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,138 @@
...
+        self.assertEqualVector(bins, [])
+        self.assertEqualVector(saliences, [])
+        self.assertEqualVector(startTimes, [])
+        # Previous tests showed small duration of 0.0058 seconds for zero or empty inputs.        
+        calculatedDuration = (2*theHopSize)/defaultSampleRate
+        self.assertAlmostEqual(duration, calculatedDuration, 8)
+
+    def testZeros(self):
+        theHopSize = 4*defaultHopSize
+        bins, saliences, startTimes, duration = PitchContours(hopSize=theHopSize)(array(zeros([2,256])), array(zeros([2,256])))
```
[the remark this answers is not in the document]
cvf-bcn-gituser: removed

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,138 @@
...
+        self.assertEqualVector(startTimes, [])
+        # Previous tests showed small duration of 0.0058 seconds for zero or empty inputs.
+        calculatedDuration = (2*theHopSize)/defaultSampleRate            
+        self.assertAlmostEqual(duration, calculatedDuration, 8)
+
+    #256 frames x small random number of zero-salience peaks at random cent bins in each frame (use a fixed random seed).
+    def testVariousZeroSaliencePeaks(self):
+        random.seed(10)
+        n = random.randrange(2, 16, 1)    
+        peakBins = array(zeros([n,256]))
```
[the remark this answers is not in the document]
cvf-bcn-gituser: I have rewritten test to have random peak an salience values for a given number of frames.

src/algorithms/tonal/pitchsaliencefunction.cpp
```diff
@@ -30,7 +30,7 @@ const char* PitchSalienceFunction::description = DOC("This algorithm computes th
 "\n"
 "This algorithm is intended to receive its \"frequencies\" and \"magnitudes\" inputs from the SpectralPeaks algorithm. The output is a vector of salience values computed for the cent bins. The 0th bin corresponds to the specified \"referenceFrequency\".\n"
 "\n"
-"When input vectors differ in size or are empty, an exception is thrown. Input vectors must contain positive frequencies and not contain negative magnitudes otherwise an exception is thrown. It is highly recommended to avoid erroneous peak duplicates (peaks of the same frequency occuring more than ones), but it is up to the user's own control and no exception will be thrown.\n"
+"If both input vectors are empty (i.e., no spectral peaks are provided), a zero salience function is returned. Input vectors must contain positive frequencies and not contain negative magnitudes otherwise an exception is thrown. It is highly recommended to avoid erroneous peak duplicates (peaks of the same frequency occuring more than ones), but it is up to the user's own control and no exception will be thrown.\n"
```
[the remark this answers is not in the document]
cvf-bcn-gituser: "If both input vectors are empty (i.e., no spectral peaks are provided), a zero salience function is returned. Input vectors must contain positive frequencies, must not contain negative magnitudes and these input vectors must be of the same size, otherwise an exception is thrown. It is highly recommended to avoid erroneous peak duplicates (peaks of the same frequency occurring more than once), but it is up to the user's own control and no exception will be thrown.\n"
"\n"

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
cvf-bcn-gituser: If binResolution is set to a high value and input size is small, a "nan" is returned in zeroth bin element.
    
    See code extract below
    
    """ FIXME

    The following test case has uncovered a situation where  divide by zero or some other 
    illegal operation has taken place because a NAN vlaue is found in output array element zero.

    def testBinResolutionTooHigh(self):        
        # Choose a sample set of frequencies and magnitude vectors of unequal length
        freq_speaks = [55] 
        mag_speaks = [1] 
        calculatedPitchSalience = PitchSalienceFunction(binResolution=55*2)(freq_speaks,mag_speaks)       
        print(calculatedPitchSalience)

        result:  calculatedPitchSalience[0] contains "nan"

    """

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
cvf-bcn-gituser: removed an replaced by realistic inputs generated by PitchSalienceFunction()

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
cvf-bcn-gituser: removed

test/src/unittests/tonal/test_pitchcontours.py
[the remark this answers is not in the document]
cvf-bcn-gituser: I dont really know what alternative you are suggesting.

In teh new testVariousSaliencePeaks() , there are random contents and sanilty checks.

test/src/unittests/tonal/test_pitchcontours.py
cvf-bcn-gituser: added

cvf-bcn-gituser:
Regarding the commented out "FIXME" Unit Test: testBinResolutionTooHigh, there is a related issue in the algorithm PitchMelodia.

If you run the following test code, the program will halt.
signal_55Hz = 0.5 * numpy.sin((array(range(40960))/44100.) * 55 * 2*math.pi)
pitch, confidence   =  PitchMelodia(binResolution=55*2)(signal_55Hz)

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,246 @@
...
+        freq_speaks = []
+        mag_speaks = []
+
+        # Randomize peak inputs
+        index = 0
+        while index < 4:
+            f1 = random.randrange(100, 500)
+            freq_speaks.append(f1) 
+            mag_speaks.append(1) 
+            index+=1
```
dbogdanov, line 107: Here, you are selecting a few random frequencies, and pass them to  `PitchSalienceFunction`. This is hard to evaluate because the values are random. Instead you could pass a single frequency (each one resulting in a straight line pitch contour, and evaluate a few of them):
- 55 Hz (corresponds to 0th bin in pitch salience function)
- 440Hz

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,246 @@
...
+        calculatedDuration =[]
+        # ... and create a Pool
+        pool = Pool();
+
+        # 2. Cut audio into frames and compute for each frame:
+        #    spectrum -> spectral peaks -> pitch salience function -> pitch salience function peaks
+        for frame in FrameGenerator(audio, frameSize=frameSize, hopSize=hopSize):
+            frame = run_windowing(frame)
+            spectrum = run_spectrum(frame)
+            peak_frequencies, peak_magnitudes = run_spectral_peaks(spectrum)
```
dbogdanov, line 202: This could be a one-liner: `run_spectral_peaks(run_spectrum(run_windowing(frame)))`

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,246 @@
...
+        contours_bins, contours_saliences, contours_start_times, duration = run_pitch_contours(
+                pool['allframes_salience_peaks_bins'],
+                pool['allframes_salience_peaks_saliences'])
+
+
+        save('calculatedcontour_bins_scale0.npy', contours_bins[0])
+        save('calculatedcontour_saliences_scale0.npy', contours_saliences[0])
+        save('calculatedcontour_bins_scale1.npy', contours_bins[1])
+        save('calculatedcontour_saliences_scale1.npy', contours_saliences[1])
+        save('calculatedcontour_startTimes_scale.npy', contours_start_times)
```
dbogdanov, line 221: We shouldn't have saves uncommented in the test.

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,246 @@
...
+        expectedSaliencesList1 = expectedSaliences1.tolist()
+        expectedStartTimesList = expectedStartTimes.tolist()
+        expectedDuration = len(audio)/(44100)
+
+        self.assertAlmostEqualVectorFixedPrecision(expectedBinsList0, contours_bins[0], 8)
+        self.assertAlmostEqualVectorFixedPrecision(expectedSaliencesList0, contours_saliences[0], 8)
+        self.assertAlmostEqualVectorFixedPrecision(expectedBinsList1, contours_bins[1], 8)
+        self.assertAlmostEqualVectorFixedPrecision(expectedSaliencesList1, contours_saliences[1], 8)
+        self.assertAlmostEqualVectorFixedPrecision(expectedStartTimesList, contours_start_times, 8)                        
+        self.assertAlmostEqual( expectedDuration, duration, 8)
```
dbogdanov, line 241: We need to test whether the correct pitch contour is output analytically, that is without referencing the output of the algorithm itself. We know how the ideal output should look like.

dbogdanov (changes requested):
Reviewed PitchContour

src/algorithms/tonal/pitchcontours.cpp
```diff
@@ -61,10 +61,16 @@ void PitchContours::compute() {
   vector<Real>& contoursStartTimes = _contoursStartTimes.get();
   Real& duration = _duration.get();
 
+  // Set duration to zero with empty input and exit.
+  if ((peakBins.size() ==0 ) || (peakSaliences.size() ==0)){
+      duration = 0;
```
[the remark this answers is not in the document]
cvf-bcn-gituser: .....Like this ?
  if (peakBins.size() ==0 ){
      contoursBins.clear();
      contoursStartTimes.clear();
      contoursSaliences.clear();
      duration = 0;
      return;
  }

test/src/unittests/tonal/test_pitchcontours.py
```diff
@@ -0,0 +1,246 @@
...
+    def testEmptyColumns(self):
+        emptyPeakBins = [[],[]]
+        emptyPeakSaliences = [[],[]]
+        theHopSize= 2*defaultHopSize
+        bins, saliences, startTimes, duration = PitchContours(hopSize=theHopSize)(emptyPeakBins, emptyPeakSaliences)
+        self.assertEqualVector(bins, [])
+        self.assertEqualVector(saliences, [])
+        self.assertEqualVector(startTimes, [])
+        calculatedDuration = (2*theHopSize)/defaultSampleRate
+        # TODO This is a huge relative error threshold. Why is that?
```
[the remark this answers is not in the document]
cvf-bcn-gituser: that comment was out of date. There is no issue here. The same formulat for duration (based on hop Size and Sample Rate) applies for empty input.

test/src/unittests/tonal/test_pitchcontours.py
cvf-bcn-gituser: To do it analytically I need to add  run_pitch_contours_melody to the algorithm chain to complete end to end test and get the rounded original pitch values from the scales ladder. I can use the original scale values as the reference

cvf-bcn-gituser:
@dbogdanov , no root cause found on duration mismatch with expected values.
Please review updated synthetic audio test. This has changed a lot to incorporate helper functions that mean no dependencies are needed on other algorithms (pitchContourMelody) and no dependency on NPY files and also to consider continuity tests.

cvf-bcn-gituser:
Too many comments, abandoning pull request.

PR #1148, opened 2021-05-18
Recommiting PitchSalienceFunction unit tests separately

cvf-bcn-gituser:
These are latest the changes from pitchsaliencefunction and pitchsaliencefunctionpeaks, separated from pitchcontour changes.


test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,378 @@
...
+        expectedBins = [270., 150. ,80. ,30.]
+        expectedPeaks =  [0.09777679, 0.07822143, 0.06257715, 0.05006172]
+        self.assertAlmostEqualVectorFixedPrecision(expectedBins, salience_peaks_bins, 6)
+        self.assertAlmostEqualVectorFixedPrecision(expectedPeaks, salience_peaks_saliences, 6)
+
+    """ FIXME
+
+    The following test case has uncovered a situation where  divide by zero or some other 
+    illegal operation has taken place because a NAN value is found in output array element zero.
+
```
dbogdanov, line 365: This is interesting.

` _binsInSemitone = floor(100.0 / _binResolution);` in the implementation will result in value 0 for `binResolution` > 100. It will then create a NAN due to division by zero. 

Another bug revealed 👏 

The solution is to prohibit values > 100 for this parameter.

dbogdanov: It would be good to cover the extreme case of `_binResolution` by a test.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def testRegression3Peaks(self):
+        # Provide 3 input peaks with a unit magnitude and
+        # validate that the output salience with previously calculated values
+        freq_speaks = [55, 100, 340]
+        mag_speaks = [1, 1, 1]
+        calculatedPitchSalience = PitchSalienceFunction()(freq_speaks,mag_speaks)
+        bins, values = PitchSalienceFunctionPeaks()(calculatedPitchSalience)
+        self.assertEqualVector(bins, [ 1., 103., 315., 195., 125.,  75.,  37.] )
+        self.assertAlmostEqualVectorFixedPrecision(values, [1.1899976, 1., 1., 0.8, 0.64000005, 0.512, 0.40960002], 6)
```
dbogdanov, line 63: This relative tolerance threshold is too huge. Let's use `assertAlmostEqualVectorFixedPrecision` instead.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def testRegression3PeaksHighBinResolution(self):
+        # Checks on config. values binResolution being too high for the input in question
+        freq_speaks = [55, 100, 340]
+        mag_speaks = [1, 1, 1]
+        calculatedPitchSalience = PitchSalienceFunction(binResolution=2400)(freq_speaks,mag_speaks)
```
dbogdanov, line 69: This would produce NAN values in PitchSalienceFunction. Is that correct?

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,160 @@
...
+        mag_speaks = [1, 1, 1]
+        calculatedPitchSalience = PitchSalienceFunction()(freq_speaks, mag_speaks)
+        self.assertRaises(RuntimeError, lambda: PitchSalienceFunctionPeaks(minFrequency=20000)(calculatedPitchSalience))
+
+    def testRegressionBadRefFreq(self):
+        # Checks on config. values referenceFrequency being too high for a given set of spectral peaks
+        freq_speaks = [55, 100, 340]
+        mag_speaks = [1, 1, 1]
+        calculatedPitchSalience = PitchSalienceFunction()(freq_speaks, mag_speaks)
+        self.assertRaises(RuntimeError, lambda: PitchSalienceFunctionPeaks(referenceFrequency=20000)(calculatedPitchSalience))
```
dbogdanov, line 93: Which algorithm is generating this error? PeakDetection of PitchSalienceFunction? We  can catch this issue in the corresponding test instead of here.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,160 @@
...
+    def testRegression3PeaksHw0UnharmonicInput(self):
+        # As in testRegression3Peaks but with Harmonic Weight = 0, non multiple freq. inputs        
+        freq_speaks = [55, 100, 340]
+        mag_speaks = [1, 1, 1]
+        expectedBins =  [0., 103., 315.]
+        expectedValues = [1., 1., 1.]
+        calculatedPitchSalience = PitchSalienceFunction(harmonicWeight=0)(freq_speaks, mag_speaks)
+        bins, values = PitchSalienceFunctionPeaks()(calculatedPitchSalience)
+        self.assertAlmostEqualVectorFixedPrecision(bins, expectedBins, 6)
+        self.assertAlmostEqualVectorFixedPrecision(values, expectedValues, 6)
```
dbogdanov, line 104: Use assertEqualVector for `expectedBins`.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
```diff
@@ -0,0 +1,160 @@
...
+    
+    def testRegression3PeaksHw0HarmonicInput(self):
+        # As in testRegression3Peaks but with Harmonic Weight = 0, multiple freq. inputs        
+        freq_speaks = [110, 220]
+        mag_speaks = [1, 1]
+        expectedBins =  [120., 240.]
+        expectedValues = [1., 1.]
+        calculatedPitchSalience = PitchSalienceFunction(harmonicWeight=0)(freq_speaks, mag_speaks)
+        bins, values = PitchSalienceFunctionPeaks()(calculatedPitchSalience)
+        self.assertAlmostEqualVectorFixedPrecision(bins, expectedBins, 6)
```
dbogdanov, line 114: Use assertEqualVector for expectedBins.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
dbogdanov: After we limit binResolution to the (0, 100] range, this test won't be necessary, but you can test with binResolution=100.

dbogdanov (changes requested):

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
cvf-bcn-gituser: change to  PitchSalienceFunction(binResolution=100

test/src/unittests/tonal/test_pitchsaliencefunction.py
cvf-bcn-gituser: need to avoid _binsInSemitone  going to zero . Update h  file

cvf-bcn-gituser:
@dbogdanov , will remove draft status as soon as I am finished with updates today

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
cvf-bcn-gituser: PitchSalienceFunctionPeaks generates the error

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
cvf-bcn-gituser: I have to use assertAlmostEqualVectorFixedPrecision() becuase the value returned is float vector for bins.
assertEqualVector doesnt work here.

test/src/unittests/tonal/test_pitchsaliencefunctionpeaks.py
cvf-bcn-gituser: PitchSalienceFunction is used to prepare the input for psfp.
it is not throwing the error.

cvf-bcn-gituser:
@dbogdanov All comments are addressed.  If any small residual issues remain, perhaps they can be addressed post merge?

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,378 @@
...
+        self.assertEqualVector(calculatedPitchSalience[4:outputLength], zeros(outputLength-4))
+
+    def testSinglePeakLowCompression(self):        
+        # Provide a single input peak with a unit magnitude at the reference frequency, and 
+        # validate that the output salience function has only one non-zero element at the first bin.             
+        freq_speaks = [55] 
+        mag_speaks = [1] 
+        outputLength  = 600        
+         # Length of the non-zero values for this Pitch Salience = 11
+        expectedPitchSalience = [1.0000000e+00, 9.7552824e-01, 9.0450847e-01, 7.9389262e-01, 6.5450847e-01,
```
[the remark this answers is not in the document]
dbogdanov: This test should be updated. Magnitude compression is `pow(magnitude, magnitudeCompression)` so the output is the same because the input magnitude is 1. Test using a smaller input magnitude value. The `expectedPitchSalience` will change.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,377 @@
...
+        calculatedPitchSalience1H = PitchSalienceFunction(numberHarmonics=1)(freq_speaks, mag_speaks)
+        self.assertEqual(len(calculatedPitchSalience1H), outputLength)               
+        calculatedPitchSalience20H = PitchSalienceFunction(numberHarmonics=20)(freq_speaks, mag_speaks)        
+        self.assertEqual(len(calculatedPitchSalience20H), outputLength)               
+        if calculatedPitchSalience1H.tolist()!=calculatedPitchSalience20H.tolist():
+        #  Return a TRUE result
+           self.assertEqual(1,1)
+        else:
+        #  Return a FALSE result       
+           self.assertEqual(1,0)
```
dbogdanov, line 137: We want to ensure that the output values are correct, not only that they are different. Therefore, instead we need to test comparing to the expected values (`expectedPitchSalience`) for both cases.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,377 @@
...
+        expectedPitchSalience = [3.0000000e+00, 2.9265847e+00, 2.7135253e+00, 2.3816779e+00, 1.9635254e+00,
+            1.5000000e+00, 1.0364745e+00, 6.1832219e-01, 2.8647447e-01, 7.3415264e-02, 1.1248198e-32]
+
+        # For 3 duplicate peaks, the expectedPitchSalience needs to be scaled by a factor of 3
+        arrayExpectedPitchSalience = 3*array(expectedPitchSalience)
+        calculatedPitchSalience = PitchSalienceFunction()(freq_speaks, mag_speaks) 
+        # The next 10 values are decreasing in magnitude   
+        self.assertAlmostEqualVectorFixedPrecision(calculatedPitchSalience[:11], expectedPitchSalience, 7)
+        # Check remaining elements are zeros
+        self.assertEqualVector(calculatedPitchSalience[12:outputLength], zeros(outputLength-(12)))
```
dbogdanov, line 154: Let's rework this. Why not append the expected amount of zerots to the `expectedPitchSalience` vector and use a single `assertAlmostEqualVectorFixedPrecision`? It will be much cleaner.
```
expectedPitchSalience += [0] * <number of expected zeros>
```

dbogdanov (changes requested):

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,394 @@
...
+        if calculatedPitchSalience1H.tolist()!=calculatedPitchSalience20H.tolist():
+        #  Return a TRUE result
+           self.assertEqual(1,1)
+        else:
+        #  Return a FALSE result       
+           self.assertEqual(1,0)        
+
+        # Detailed contents check on returned values (regression check
+        self.assertAlmostEqualVectorFixedPrecision(calculatedPitchSalience1H[:11], expectedPitchSalienceH1, 7)
+        self.assertAlmostEqualVectorFixedPrecision(calculatedPitchSalience20H[:11], expectedPitchSalienceH20, 7)
```
dbogdanov, line 154: Are the rest of the values zeros? Test comparing the entire 600-value vectors similarly to other tests. Otherwise, it is misleading, because it is not clear what happens to the rest of the values.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,394 @@
...
+        self.assertEqual(len(calculatedPitchSalience1H), outputLength)
+        calculatedPitchSalience20H = PitchSalienceFunction(numberHarmonics=20)(freq_speaks, mag_speaks)
+        self.assertEqual(len(calculatedPitchSalience20H), outputLength)
+
+        if calculatedPitchSalience1H.tolist()!=calculatedPitchSalience20H.tolist():
+        #  Return a TRUE result
+           self.assertEqual(1,1)
+        else:
+        #  Return a FALSE result       
+           self.assertEqual(1,0)
```
dbogdanov, line 150: We don't really need this because we have hardcoded `expectedPitchSalienceH1` and `calculatedPitchSalience20H` we compare against, which are different.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,378 @@
...
+        expectedPitchSalience = load(join(filedir(), 'pitchsalience/calculatedPitchSalience_vignesh.npy'))
+        expectedPitchSalienceList = expectedPitchSalience.tolist()
+
+        index=0
+        for frame in FrameGenerator(audio, frameSize=frameSize, hopSize=hopSize):
+            # Tolerance check to 6 decimal palces (for Linux Subsyst. compatibility.
+            self.assertAlmostEqualVectorFixedPrecision(expectedPitchSalienceList[index], calculatedPitchSalience[index], 6)
+            index+=1        
+
+    def testRegressionSyntheticInput(self):
```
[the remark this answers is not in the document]
dbogdanov: Move this test to PitchSalienceFunctionPeaks, it does not belong here.

test/src/unittests/tonal/test_pitchsaliencefunction.py
```diff
@@ -0,0 +1,394 @@
...
+
+        calculatedPitchSalience = PitchSalienceFunction(magnitudeThreshold=0)(freq_speaks, mag_speaks)
+        # With Mag. Threshold  = 0 outputs are zero
+        self.assertEqual(len(calculatedPitchSalience), outputLength)       
+        self.assertEqualVector(calculatedPitchSalience, zeros(600))        
+        """ FIXME
+        The intention here is to test if some peaks are correctly discarded, however, this test does not achieve this goal because there is only one input spectral peak.    
+        The magnitudeThreshold parameter defines the maximum allowed difference from the highest peak in dBs, so the input list of peaks should be of different magnitudes to see it working.
+        For example, to make it simple, we can add a few peaks with lower magnitudes than 1 and use magnitudeThreshold = 0. Then only the peak with the magnitude 1 will be selected and the current expectedPitchSalience will be still a correct expected test output
+        """
```
dbogdanov, line 121: This has to be addressed in the test. The current test does not make much utility.

dbogdanov (changes requested):

PR #1158, opened 2021-05-27
Indian tonic: Fixed bug relating to use cases that produced empty peaks

cvf-bcn-gituser:
While testing tonicindianartmusic.cpp, I found a some cases where segmentation faults occur. Initially I thought that it was related to the empty audio, but on closer examination, in a test case where you send in a pure sine wave with a frequency below the configured minTonicFrequency, can produce a set of empty peaks and consequently a segmentation fault.

I added 8 test cases. They may not cover all config combinations, but catch the typical illegal ones (e.g.  min and max Tonics conflicting), but its a start to at least give good coverage to the change in the CPP file.

### DIFF
--- a/src/algorithms/tonal/tonicindianartmusic.cpp
+++ b/src/algorithms/tonal/tonicindianartmusic.cpp
@@ -29,7 +29,7 @@ namespace standard {
 
 const char* TonicIndianArtMusic::name = "TonicIndianArtMusic";
 const char* TonicIndianArtMusic::category = "Tonal";
-const char* TonicIndianArtMusic::description = DOC("This algorithm estimates the tonic frequency of the lead artist in Indian art music. It uses multipitch representation of the audio signal (pitch salience) to compute a histogram using which the tonic is identified as one of its peak. The decision is made based on the distance between the prominent peaks, the classification is done using a decision tree.\n"
+const char* TonicIndianArtMusic::description = DOC("This algorithm estimates the tonic frequency of the lead artist in Indian art music. It uses multipitch representation of the audio signal (pitch salience) to compute a histogram using which the tonic is identified as one of its peak. The decision is made based on the distance between the prominent peaks, the classification is done using a decision tree. An empty input signal will throw an exception. An exception will also be thrown if no predominant pitch salience peaks are detected within the maxTonicFrequency to minTonicFrequency range. \n"
 "\n"
 "References:\n"
 "  [1] J. Salamon, S. Gulati, and X. Serra, \"A Multipitch Approach to Tonic\n"
@@ -198,14 +198,12 @@ void TonicIndianArtMusic::compute() {
   /*implementing the decision tree*/
 
   // Prevent segmentation fault
-  if (peak_locs.size() == 0) { 
+  if (peak_locs.size() == 0){ 
     throw EssentiaException("TonicIndianArtMusic: No peak locations"); 
   }
 
-    // Prevent segmentation fault
-  if (peak_amps.size() == 0) { 
-    throw EssentiaException("TonicIndianArtMusic: No peak amplitudes"); 
-  }
+  // Peak detection should guarantee the equal number of both types of value.
+  assert(peak_locs.size() == peak_amps.size());
   
   Real highest_peak_loc = peak_locs[0];
   Real f2 =peak_locs[1] - highest_peak_loc;
--- a/src/algorithms/tonal/tonicindianartmusic.cpp
+++ b/src/algorithms/tonal/tonicindianartmusic.cpp
@@ -95,9 +95,18 @@ void TonicIndianArtMusic::configure() {
 
 
 void TonicIndianArtMusic::compute() {
+  
+  printf("DBG0");
+  std::cout<< "DEBUGGING"<< std::endl;
   const vector<Real>& signal = _signal.get();
   Real& tonic = _tonic.get();
+  printf("DBG1");
+  // Prevent segmentation fault
+  if (signal.size() == 0) { 
+    throw EssentiaException("TonicIndianArtMusic: Empty Audio passed"); 
+  }
 
+  printf("DBG2");
   // Pre-processing
   vector<Real> frame;
   _frameCutter->input("signal").set(signal);
--- a/src/algorithms/tonal/tonicindianartmusic.cpp
+++ b/src/algorithms/tonal/tonicindianartmusic.cpp
@@ -96,31 +96,14 @@ void TonicIndianArtMusic::configure() {
 
 void TonicIndianArtMusic::compute() {
   
-  printf("DBG0 \n");
-  std::cout<< "DEBUGGING"<< std::endl;
-
-
-
-  printf("DBG0b \n");
   const vector<Real>& signal = _signal.get();
-    // Prevent segmentation fault
+  // Prevent segmentation fault
   if (signal.size() == 0) { 
     throw EssentiaException("TonicIndianArtMusic: Empty Audio passed"); 
   }
-  printf("DBG0c \n");  
-
 
   Real& tonic = _tonic.get();
-  printf("DBG1 \n");
-  if (tonic == 0) { 
-    printf("TonicIzero \n"); 
-  }
-  else
-  {
-     printf("TonicNOJNzero \n"); 
-  }
-  print("CRASH HERE!!!!!! for    x = 0.5 * numpy.sin((array(range(signalSize))/44100.) * 99.1* 2*math.pi)")
-  printf("DBG2");
+
   // Pre-processing
   vector<Real> frame;
   _frameCutter->input("signal").set(signal);
@@ -142,7 +125,6 @@ void TonicIndianArtMusic::compute() {
   _spectralPeaks->output("magnitudes").set(frameMagnitudes);
 
 
-  printf("DBG3");
   // Pitch salience contours
   vector<Real> frameSalience;
   _pitchSalienceFunction->input("frequencies").set(frameFrequencies);
@@ -155,8 +137,6 @@ void TonicIndianArtMusic::compute() {
   _pitchSalienceFunctionPeaks->output("salienceBins").set(frameSalienceBins);
   _pitchSalienceFunctionPeaks->output("salienceValues").set(frameSalienceValues);
 
-
-  printf("DBG4");
   // histogram computation
   vector<Real> histogram;
   histogram.resize(_numberBins);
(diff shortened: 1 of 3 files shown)


test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -0,0 +1,119 @@
...
+        self.assertConfigureFails(TonicIndianArtMusic(), { 'referenceFrequency': -1 })
+        self.assertConfigureFails(TonicIndianArtMusic(), { 'sampleRate':   -1 })
+
+    def testZeros(self):
+        self.assertRaises(RuntimeError, lambda: TonicIndianArtMusic()(zeros(1024)))   
+
+    def testOnes(self):
+        referenceTonic =108.86099243164062     
+        tonic   = TonicIndianArtMusic()(ones(1024))
+        self.assertAlmostEqual(referenceTonic, tonic, 8)
```
dbogdanov, line 51: See the definition of `assertAlmostEqual` in [essentia_test.py](https://github.com/MTG/essentia/blob/master/test/src/unittests/essentia_test.py#L125): 
> def assertAlmostEqual(self, found, expected, precision = 1e-7):

so the output and expected output values are in a wrong order. 
Also, why the margin of error (8) is so huge? Does it even make sense?

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -0,0 +1,119 @@
...
+        self.assertConfigureFails(TonicIndianArtMusic(), { 'numberHarmonics': 0 })
+        self.assertConfigureFails(TonicIndianArtMusic(), { 'numberHarmonics': -1 })        
+        self.assertConfigureFails(TonicIndianArtMusic(), { 'numberSaliencePeaks': 0})
+        self.assertConfigureFails(TonicIndianArtMusic(), { 'referenceFrequency': -1 })
+        self.assertConfigureFails(TonicIndianArtMusic(), { 'sampleRate':   -1 })
+
+    def testZeros(self):
+        self.assertRaises(RuntimeError, lambda: TonicIndianArtMusic()(zeros(1024)))   
+
+    def testOnes(self):
```
dbogdanov, line 48: We can test a constant silence input instead as it makes more sense as a possible real-world input.

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -0,0 +1,119 @@
...
+        referenceTonic =108.86099243164062     
+        tonic   = TonicIndianArtMusic()(ones(1024))
+        self.assertAlmostEqual(referenceTonic, tonic, 8)
+
+    def testNegativeInput(self):
+        tonic = TonicIndianArtMusic()([-1]*1024)
+        referenceTonic =108.86099243164062     
+        tonic   = TonicIndianArtMusic()([-1]*1024)
+        self.assertAlmostEqual(referenceTonic, tonic, 8)
```
dbogdanov, line 57: This is another constant input test that doesn't make much sense in the context of real-world audio signal.

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def testRegression(self):
+        audio = MonoLoader(filename = join(testdata.audio_dir, 'recorded/vignesh.wav'),
+                            sampleRate = 44100)()
+        referenceTonic = 102.74                                       
+        tonic = TonicIndianArtMusic()(audio)
+        self.assertAlmostEqual(referenceTonic, tonic, 6)
```
dbogdanov, line 64: Same here, wrong order for test variables, and very large error margin. I wonder why is that. What is the output tonic `value`?

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def testMinMaxMismatch(self):
+        frameSize = 2048
+        signalSize = 15 * frameSize
+        x = 0.5 * numpy.sin((array(range(signalSize))/44100.) * 99.1* 2*math.pi)
```
dbogdanov, line 69: Avoid redundant details simplify the code for this test. Here we only want to know that there is an exception thrown when minTonicFrequency is larger than maxTonicFrequency. The input signal values are irrelevant.

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -0,0 +1,119 @@
...
+        self.assertRaises(RuntimeError, lambda: TonicIndianArtMusic(minTonicFrequency=190, maxTonicFrequency=11)(x))
+
+    def testBelowMinimumTonic(self):
+        # generate test signal 99 Hz, and put minFreq as 100 Hz in the TonicIndianArtMusic
+        defaultSampleRate = 44100
+        frameSize = 2048
```
dbogdanov, line 75: Variables unused

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -0,0 +1,119 @@
...
+        tonic = TonicIndianArtMusic()(audio)
+        self.assertAlmostEqual(referenceTonic, tonic, 6)
+
+    def testMinMaxMismatch(self):
+        frameSize = 2048
+        signalSize = 15 * frameSize
+        x = 0.5 * numpy.sin((array(range(signalSize))/44100.) * 99.1* 2*math.pi)
+        self.assertRaises(RuntimeError, lambda: TonicIndianArtMusic(minTonicFrequency=190, maxTonicFrequency=11)(x))
+
+    def testBelowMinimumTonic(self):
```
dbogdanov, line 72: What about a similar test for above the maximum tonic?

dbogdanov (changes requested):
Great catch for the memory bug. 👍

cvf-bcn-gituser:
@dbogdanov.  
I think I found another bug in Indian Tonic , but I dont yet have the root cause. 
See comment on last commit:  "Added FIXME to testOnes, testRegression  where tonic returned is not always the same".
Will commit again when I find root cause.

src/algorithms/tonal/tonicindianartmusic.cpp
```diff
@@ -198,14 +198,12 @@ void TonicIndianArtMusic::compute() {
   /*implementing the decision tree*/
 
   // Prevent segmentation fault
-  if (peak_locs.size() == 0) { 
+  if (peak_locs.size() == 0){
```
dbogdanov, line 201: This error message is not very informative to the user. Alternative:
> TonicIndianArtMusic: No predominant pitch salience peaks found in the specified tonic frequency region [minTonicFrequency, maxTonicFrequency]".

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -33,66 +34,88 @@ def testInvalidParam(self):
...
-        self.assertRaises(RuntimeError, lambda: TonicIndianArtMusic()(zeros(1024)))   
+    def testEmpty(self):
+        self.assertRaises(RuntimeError, lambda: TonicIndianArtMusic()([]))   
 
-    def testOnes(self):
-        referenceTonic =108.86099243164062     
-        tonic   = TonicIndianArtMusic()(ones(1024))
-        self.assertAlmostEqual(referenceTonic, tonic, 8)
+    def testSilence(self):
+        silence = np.zeros(int(np.abs(np.random.randn()) * 30. * 44100))
```
dbogdanov, line 52: We can use a fixed zero input, instead of a random size vector.

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -33,66 +34,88 @@ def testInvalidParam(self):
...
+        self.assertRaises(RuntimeError, lambda: TonicIndianArtMusic()(silence))
 
-    def testNegativeInput(self):
-        tonic = TonicIndianArtMusic()([-1]*1024)
-        referenceTonic =108.86099243164062     
-        tonic   = TonicIndianArtMusic()([-1]*1024)
-        self.assertAlmostEqual(referenceTonic, tonic, 8)
+    def testOnes(self):
+        # FIXME: The tonic value 117.36 is not returned everytime. Sometimes 55 is returned.
```
dbogdanov, line 56: What is the source for deciding the correct `referenceTonic` value? The constant input signal doesn't contain a workable frequency spectrum, so i don't see the point of this test.

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -33,66 +34,88 @@ def testInvalidParam(self):
...
+        tonic   = TonicIndianArtMusic()(ones(4096))
+        self.assertAlmostEqual(tonic, referenceTonic, 5)
 
     def testRegression(self):
         audio = MonoLoader(filename = join(testdata.audio_dir, 'recorded/vignesh.wav'),
                             sampleRate = 44100)()
         referenceTonic = 102.74                                       
         tonic = TonicIndianArtMusic()(audio)
-        self.assertAlmostEqual(referenceTonic, tonic, 6)
+        self.assertAlmostEqual( tonic, referenceTonic, 6)
```
dbogdanov, line 67: `6` is a very large deviation, how do you decide the value? what are the estimated vs reference values?

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -33,66 +34,88 @@ def testInvalidParam(self):
...
+        start_zero = np.zeros(int(np.abs(np.random.randn()) * 30. * 44100))
+        end_zero = np.zeros(int(np.abs(np.random.randn()) * 30. * 44100))
+        # Check result is the same with appended silences
+        real_audio = np.hstack([start_zero, audio, end_zero])
+        tonic = TonicIndianArtMusic()(real_audio)
+        # FIXME: tonic is sometimes 102.74112701416016 and other times  154.9361114501953
+        self.assertAlmostEqual(tonic, referenceTonic, 5)
```
dbogdanov, line 74: Same here

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -33,66 +34,88 @@ def testInvalidParam(self):
...
+
+    def testWhiteNoise(self):
+        from numpy.random import uniform
+        sig = array(uniform(size=10000))
+        tonic  = TonicIndianArtMusic()(sig)
```
dbogdanov, line 79: TonicIndianArtMusic expects an input audio signal, so this test is invalid (the uniform distribution is for the frequency domain, not the time domain).

On the other side, not sure it is really necessary either.

test/src/unittests/tonal/test_tonicindianartmusic.py
```diff
@@ -33,66 +34,88 @@ def testInvalidParam(self):
...
+        x = 0.5 * numpy.sin((array(range(signalSize))/44100.) * 101* 2*math.pi)        
+        self.assertRaises(RuntimeError, lambda: TonicIndianArtMusic(minTonicFrequency=99,maxTonicFrequency=100)(x))
+ 
     def testRegressionSyntheticSignal(self):
-
-        # generate test signal concatenating major scale notes.
-        defaultSampleRate = 44100
+        # generate a test signal concatenating different frequencies
         frameSize = 2048
         signalSize = 15 * frameSize
```
dbogdanov: why not 
```
signalSize = 2048 * 15
```
The `frameSize` variable doen't have any use later on.

dbogdanov (changes requested):

src/algorithms/tonal/tonicindianartmusic.cpp
cvf-bcn-gituser: updated

test/src/unittests/tonal/test_tonicindianartmusic.py
cvf-bcn-gituser: use ground truth from data set. and check its OK without silences

test/src/unittests/tonal/test_tonicindianartmusic.py
cvf-bcn-gituser: removed

PR #1161, opened 2021-06-03
Support Tensorflow models in SavedModel format

palonso:
This PR gives `TensorflowPredict` and derivate algorithms the capability to run Tensorflow SavedModels.
When both `savedModel` and `graphFilename` are specified, the first one is preferred

Additionally, it modifies `TensorflowPredict`'s behavior so that the configuration is skipped  when `graphFilename` or/and `savedModel` are set to an empty string ("").
The main benefit is that algorithms depending on `TensorflowPredict` can default `graphFilename` and `savedModel` to an empty string and leverage the skip-configuration logic implemented on `TensorflowPredict`.
This also simplifies the logic required to choose between the `graphFilename` and `savedModel` when both are specified.

### DIFF
--- a/src/algorithms/machinelearning/tensorflowpredict.cpp
+++ b/src/algorithms/machinelearning/tensorflowpredict.cpp
@@ -49,8 +49,11 @@ static void DeallocateBuffer(void* data, size_t) {
 
 void TensorflowPredict::configure() {
   // If no file has been specified, do not do anything
-  if (!parameter("graphFilename").isConfigured()) return;
-  _graphFilename = parameter("graphFilename").toString();
+  if (parameter("savedModel").isConfigured()) _savedModel = parameter("savedModel").toString();
+  else if (parameter("graphFilename").isConfigured()) _graphFilename = parameter("graphFilename").toString();
+  else return;
+
+  _tags = parameter("tags").toVectorString();
 
   _inputNames = parameter("inputs").toVectorString();
   _outputNames = parameter("outputs").toVectorString();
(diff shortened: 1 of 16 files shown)


src/algorithms/machinelearning/tensorflowpredict.cpp
```diff
@@ -48,9 +48,16 @@ static void DeallocateBuffer(void* data, size_t) {
...
-  if (!parameter("graphFilename").isConfigured()) return;
-  _graphFilename = parameter("graphFilename").toString();
+  _savedModel = "";
+  _graphFilename = "";
+
+  if (parameter("savedModel").isConfigured()) _savedModel = parameter("savedModel").toString();
+  if (parameter("graphFilename").isConfigured()) _graphFilename = parameter("graphFilename").toString();
+
+  // Do not do anything if we did not get a non-empty model name.
+  if ((_savedModel.empty()) and (_graphFilename.empty())) return;
```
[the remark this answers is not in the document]
dbogdanov: May be not if we E_INFO when the model is specified and loaded instead.

src/algorithms/machinelearning/tensorflowpredict.h
```diff
@@ -87,10 +92,16 @@ class TensorflowPredict : public Algorithm {
...
     TF_DeleteGraph(_graph);
+    TF_DeleteBuffer(_runOptions);
   }
 
   void declareParameters() {
-    declareParameter("graphFilename", "the name of the file from which to read the Tensorflow graph", "", Parameter::STRING);
+    const char* defaultTagsC[] = { "serve" };
+    std::vector<std::string> defaultTags = arrayToVector<std::string>(defaultTagsC);
+
+    declareParameter("graphFilename", "the name of the file from which to read the TensorFlow graph", "", Parameter::STRING);
```
dbogdanov, line 102: "read" -> "load" ?

test/src/unittests/machinelearning/test_tensorflowpredict.py
```diff
@@ -59,19 +58,76 @@ def testRegression(self):
...
+            'savedModel': join(testdata.models_dir, 'vgg', 'vgg4'),
+            'inputs': ['model/Placeholder'],
+            'outputs': ['model/Softmax'],
+            'isTraining': False,
+            'isTrainingName': 'model/Placeholder_1',
+        }
+
+        self.regression(parameters)
+
+    def testSavedModelPreferecce(self):
```
dbogdanov, line 90: "Preference". But a more informative test name is better. `testSavedModelOverridesGraphFilename` ?

test/src/unittests/machinelearning/test_tensorflowpredict.py
```diff
@@ -59,19 +58,76 @@ def testRegression(self):
...
+            'savedModel': join(testdata.models_dir, 'vgg', 'vgg4'),
+            'inputs': ['model/Placeholder'],
+            'outputs': ['model/Softmax'],
+            'isTraining': False,
+            'isTrainingName': 'model/Placeholder_1',
+        }
+
+        self.regression(parameters)
+
+    def testEmptyModelName(self):
```
dbogdanov, line 104: There is no assert in this test.

test/src/unittests/machinelearning/test_tensorflowpredictmusicnn.py
```diff
@@ -35,8 +35,43 @@ def testRegression(self):
...
+        filename = join(testdata.audio_dir, 'recorded', 'vignesh.wav')
+        model = join(testdata.models_dir, 'musicnn', 'genre_dortmund_musicnn_msd')
+
+        audio = MonoLoader(filename=filename, sampleRate=16000)()
+        found = TensorflowPredictMusiCNN(savedModel=model)(audio)[0]
+
+        self.assertAlmostEqualVector(found, expected, 1e-5)
+
+    def testEmptyModelName(self):
+        # With empty or undefined model names the algorithm should skip the configuration without errors.
```
dbogdanov, line 53: Same here

test/src/unittests/machinelearning/test_tensorflowpredicttempocnn.py
```diff
@@ -115,6 +120,19 @@ def testInvalidParam(self):
         self.assertConfigureFails(TensorflowPredictTempoCNN(), {'graphFilename': 'non_existing_model.pb',
                                                                 'output': 'non_existing'})
 
+    def testEmptyModelName(self):
```
dbogdanov, line 123: Same here

test/src/unittests/machinelearning/test_tensorflowpredictvggish.py
```diff
@@ -40,6 +40,45 @@ def testRegression(self):
...
+        audio = MonoLoader(filename=filename, sampleRate=16000)()
+        found = TensorflowPredictVGGish(savedModel=model, patchHopSize=0)(audio)
+        found = numpy.mean(found, axis=0)
+
+        # Setting a high tolerance value due to the mismatch between the
+        # original and replicated features. However, they are close enough to
+        # make valid predictions.
+        self.assertAlmostEqualVector(found, expected, 1e-2)
+
+    def testEmptyModelName(self):
```
dbogdanov, line 58: Same here

dbogdanov (changes requested):

src/algorithms/machinelearning/tensorflowpredict.cpp
```diff
@@ -48,9 +48,16 @@ static void DeallocateBuffer(void* data, size_t) {
 
 
 void TensorflowPredict::configure() {
-  // If no file has been specified, do not do anything
-  if (!parameter("graphFilename").isConfigured()) return;
-  _graphFilename = parameter("graphFilename").toString();
+  _savedModel = "";
+  _graphFilename = "";
+
+  if (parameter("savedModel").isConfigured()) _savedModel = parameter("savedModel").toString();
```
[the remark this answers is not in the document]
palonso: Good point. I've tried it and it passed all the tests.  

I'll change `graphFilename` and `savedModel`'s default in `TensorflowPredict` to "" so we can simplify the logic to skip configuration.

test/src/unittests/essentia_test.py
```diff
@@ -187,6 +187,18 @@ def assertConfigureFails(self, algo, params):
         conf = lambda: algo.configure(**params)
         self.assertRaises(EssentiaException, conf)
 
+    def assertConfigureSuccess(self, algo, params):
+        try:
+            algo.configure(**params)
+        except EssentiaException:
```
dbogdanov, line 193: what if `configure` fails for some other exception type, not being `EssentiaException`?

test/src/unittests/essentia_test.py
```diff
@@ -187,6 +187,18 @@ def assertConfigureFails(self, algo, params):
...
+            self.fail()
+
+    def assertComputeSuccess(self, algo, params):
+        try:
+            algo.compute(**params)
+        except EssentiaException:
```
dbogdanov, line 199: Same here

dbogdanov (changes requested):

test/src/unittests/essentia_test.py
palonso: In that case, the test will return `error` instead of `fail`.
As I understand it, the goal of this assertion is to test if a particular case skips an exception that is the general rule.
The way I think it is:
pass -> ok
fail -> we caught the error but we didn't consider that this particular case should be skipped
error -> we did not catch the error

Do you think we should except any exception instead?

dbogdanov: Yes, I think that is fine, as the case of ERROR will help to know that we did not catch some exception inside `configure`.

PR #1166, opened 2021-06-14
Two draft unit test scripts (on "frame2real" and  "tensor2vectorreal") from the Auxillary section of UTs

cvf-bcn-gituser:
Documenting the unchecked work started on the following Unit Test Files:
test_frametoreal.py 
test_tensortovectorreal.py (based on the "mirror" version)


cvf-bcn-gituser:
@dbogdanov , will add tests similar to the ones in test_framecutter.py

PR #1173, opened 2021-07-21
Monolithic build of Tensorflow 2.5

palonso:
Update the `build_tensorflow.sh` and TensorFlow dependency version.

Now TensorFlow is built with Bazel on its monolithic configuration. This is a shared library with an almost static behavior.
Critically, the monolithic version of `libtensorflow.so` does not depend on `framework_tensorflow.so` which contains the capability to register ops from outside the library and was the cause of the clash with the TensorFlow wheel.

Also, the `essentia-tensorflow` wheels are linked against our TensorFlow build, so TensorFlow is not a Python setup dependency anymore.

This approach gives us the capability to make `essentia-tensorflow` wheels up to the latest Python version (3.9 at the moment).

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -58,7 +58,7 @@ Note that [Bazel](https://bazel.build/) has to be installed to build TensorFlow.
 For simplicity, we recommend to get Bazel through [Bazelisk](https://docs.bazel.build/versions/main/install-bazelisk.html).
 To build TensorFlow with GPU support [CUDA](https://developer.nvidia.com/cuda-toolkit) and [cuDNN](https://developer.nvidia.com/cudnn) have to be installed.
 Otherwise, GPU support can be deactivated by setting ```export TF_NEED_CUDA=0``` before building TensorFlow.
-More TensorFlow related configuration is available in the [build configuration file](https://github.com/MTG/essentia/blob/master/packaging/build_config.sh).
+More TensorFlow-related configuration is available in the [build configuration file](https://github.com/MTG/essentia/blob/master/packaging/build_config.sh).
 
 Alternatively, you can build each dependency apart running the corresponding scripts inside ```packaging/debian_3rdparty``` folder:
 ```
--- a/FAQ.md
+++ b/FAQ.md
@@ -54,6 +54,12 @@ Use ```--with-gaia``` flag to include Gaia.
 
 Use ```--with-tensorflow``` flag to include TensorFlow.
 
+Note that [Bazel](https://bazel.build/) has to be installed to build TensorFlow.
+For simplicity, we recommend to get Bazel through [Bazelisk](https://docs.bazel.build/versions/main/install-bazelisk.html).
+To build TensorFlow with GPU support [CUDA](https://developer.nvidia.com/cuda-toolkit) and [cuDNN](https://developer.nvidia.com/cudnn) have to be installed.
+Otherwise, GPU support can be deactivated by setting ```export TF_NEED_CUDA=0``` before building TensorFlow.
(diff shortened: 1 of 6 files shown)


FAQ.md
```diff
@@ -54,6 +54,12 @@ Use ```--with-gaia``` flag to include Gaia.
 
 Use ```--with-tensorflow``` flag to include TensorFlow.
 
+Note that [Bazel](https://bazel.build/) has to be installed to build TensorFlow.
+For simplicity, we recommend to get Bazel through [Bazelisk](https://docs.bazel.build/versions/main/install-bazelisk.html).
+To build TensorFlow with GPU support [CUDA](https://developer.nvidia.com/cuda-toolkit) and [cuDNN](https://developer.nvidia.com/cudnn) have to be installed.
+Otherwise, GPU support can be deactivated by setting ```export TF_NEED_CUDA=0``` before building TensorFlow.
+More TensorFlow related configuration is available in the [build configuration file](https://github.com/MTG/essentia/blob/master/packaging/build_config.sh).
```
dbogdanov, line 61: "TensorFlow-releated"

packaging/debian_3rdparty/build_tensorflow.sh
```diff
@@ -10,75 +10,30 @@ cd tmp
 
 echo "Building Tensorflow $TENSORFLOW_VERSION"
 
+# NumPy is required by Bazel
+${PYTHON_BIN_PATH} -m pip install numpy
```
dbogdanov, line 14: Why the `numpy` Python package is required? 

Btw, we already demand `numpy` dependency in https://essentia.upf.edu/installing.html#installing-dependencies-on-linux, so we can expect a user to install them before running build scripts.

travis/build_wheels.sh
```diff
@@ -47,60 +39,35 @@ cd -
...
-            NUMPY_VERSION=1.14.5
-        elif [[ $PYBIN == *"cp36"* ]]; then
-            NUMPY_VERSION=1.11.3
-        elif [[ $PYBIN == *"cp34"* ]] || [[ $PYBIN == *"cp35"* ]]; then
-            NUMPY_VERSION=1.9.3
-        fi
-
-        "${PYBIN}/pip" install numpy==$NUMPY_VERSION
+    # Python 2.7
+    NUMPY_VERSION=1.8.2
```
dbogdanov, line 49: Redundant code? We don't support Python 2.

travis/build_wheels.sh
```diff
@@ -47,60 +39,35 @@ cd -
...
+    # Python 3.x
+    if [[ $PYBIN == *"cp39"* ]]; then
+        NUMPY_VERSION=1.19.3
+    elif [[ $PYBIN == *"cp38"* ]]; then
+        NUMPY_VERSION=1.17.4
+    elif [[ $PYBIN == *"cp37"* ]]; then
+        NUMPY_VERSION=1.14.5
+    elif [[ $PYBIN == *"cp36"* ]]; then
+        NUMPY_VERSION=1.11.3
+    elif [[ $PYBIN == *"cp34"* ]] || [[ $PYBIN == *"cp35"* ]]; then
```
dbogdanov, line 60: We discard these outdated Python versions, so should this be commented? It may be good to keep the required NUMPY_VERSION for them documented.

dbogdanov (changes requested):

travis/build_wheels.sh
palonso: Ok, I will comment  out code related to outdated Python versions.

dbogdanov (approved):

PR #1174, opened 2021-08-10
Add CREPE models

palonso:
This PR adds
- `CREPE` and `TensorflowPredictCREPE`
- Related unit tests
- Fixes a small bug in `VectorRealToTensor`


src/algorithms/tonal/crepe.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  for (int i = 0; i < timestamps; i++) {
+    // Get the timestamp of each pitch prediction.
+    time[i] = i * _hopSize / 1000.0;
+
+    // Get the confidences (max Activation per timestamp).
+    confidence[i] = activations[i][argmax(activations[i])];
+
+    // Cents to frequencies.
+    if (!isnan(cents[i])) {
```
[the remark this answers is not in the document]
palonso: `nan` can happen when:
- a frame is unvoiced, so all the activations sum to zero and there is a division by zero computing the cents.
- the activation matrix itself contain `nans`. I guess this could happen, but I have never seen it.

I think it's better to handle the unvoiced timestamps in the Herz level because:
1. while we can use 0Hz to represent unvoiced timestamps, 0 cents actually refers to the lowest pitch the model is able to predict.
2. This implementation is parallel to the original Python implementation, which can be useful for future debugging.

src/algorithms/tonal/crepe.cpp
```diff
@@ -0,0 +1,131 @@
...
+      frequency[i] = 10.0 * pow(2.0, (cents[i] / 1200.0));
+    }
+  }
+}
+
+vector<Real> CREPE::toLocalAverageCents(vector<vector<Real> > &activations) {
+  int timestamps = activations.size();
```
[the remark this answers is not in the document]
palonso: I've never seen this, nor I could find issues related to `nans` in the official CREPE reposition, which also applies post-processing on top of the activation matrix generated by the model

src/algorithms/tonal/crepe.cpp
```diff
@@ -0,0 +1,131 @@
...
+  vector<Real> cents(timestamps, 0);
+
+  int center, start, end;
+  Real productSum, weightSum;
+  vector<Real> salience, toCents;
+  for (int i = 0; i < timestamps; i++) {
+    center = argmax(activations[i]);
```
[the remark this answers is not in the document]
palonso: Nice observation!
Our argmax implementation relies on [max_element](https://www.cplusplus.com/reference/algorithm/max_element/), which as NumPy's [argmax](https://numpy.org/doc/stable/reference/generated/numpy.argmax.html) returns the first element in case of multiple candidates.
Then we should decide if matching the original implantation is enough, or if we should look for a method to choose a candidate if there is more than one.

I couldn't find any issue related to this in the [original repo](https://github.com/marl/crepe).

src/algorithms/tonal/crepe.cpp
```diff
@@ -0,0 +1,131 @@
...
+  for (int i = 0; i < timestamps; i++) {
+    // Get the timestamp of each pitch prediction.
+    time[i] = i * _hopSize / 1000.0;
+
+    // Get the confidences (max Activation per timestamp).
+    confidence[i] = activations[i][argmax(activations[i])];
+
+    // Cents to frequencies.
+    if (!isnan(cents[i])) {
+      frequency[i] = 10.0 * pow(2.0, (cents[i] / 1200.0));
```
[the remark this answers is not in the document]
palonso: We have `Hz2Cents` but not `Cents2Hz`. I was trying to implement it, but it looks like `Hz2Cents` is not the inverse of the operation we need so I have to investigate this to clarify what would be the best option for `essentiamath.h`

For now I'll leave the computation inside the algorithm.

src/algorithms/tonal/crepe.cpp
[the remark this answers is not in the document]
palonso: A new observation:
In the [original implementation](https://github.com/marl/crepe/blob/a666a03011a9cdab70e9abd0b5009ad60c5f8926/crepe/core.py#L207), frames are normalized in a way that constant frames (e.g., zeros) produce `NaNs`: `( (x - mean) / stddev )`

These `NaNs` are propagated to the activation matrix, and are then interpreted as frequencies equal to 0 according to the logic that we implemented on line 104.

CREPE has an open [PR](https://github.com/marl/crepe/pull/76/) proposing to use `EPS` to prevent division by 0, but the authors haven't made any comment about it.
I have tried the PR, and it effectively prevents the `NaNs` to happen and the CNN return incorrect pitch values with low confidence (activation) values. I guess the authors may prefer the current behavior, as it returns clean zeros for constant frames. 

Essentia has two mechanisms to avoid this:
- By default `FrameCutter` adds noise to silence frames.
- `TensorNormalice` does not divide by the stddev when it is equal to zero

For now, I decided to deactivate both mechanisms so that our normalization also returns `NaNs` on constant frames, so that we mimic the original implementation.

src/algorithms/tonal/crepe.cpp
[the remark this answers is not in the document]
dbogdanov: We can be as consistent with the original implementation as possible. But may be it is worth mentioning this issue somewhere, at least in the code comments.

src/algorithms/tonal/crepe.cpp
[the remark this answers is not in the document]
dbogdanov: On the other side, `NaN` is not part of JSON and we need to double-check if it is possible to store such an output with our `YamlOutput` algorithm (and we are missing a unit test for that in `test_yamloutput.py`).

doc/sphinxdoc/research_papers.md
```diff
@@ -80,7 +80,11 @@ focused on weblogs. In 7th International Conference on Music Information Retriev
 
 ## Emotion detection
 
+<<<<<<< HEAD
```
dbogdanov: Some merge conflict here.

src/algorithms/machinelearning/tensorflowpredict.h
```diff
@@ -53,6 +53,11 @@ class TensorflowPredict : public Algorithm {
   TF_SessionOptions* _sessionOptions;
   TF_Session* _session;
 
+  std::string _savedModel;
+  std::vector<std::string> _tags;
+  TF_Buffer* _runOptions;
+  TF_Buffer* _metaGraphDef;
```
dbogdanov: I can't see this variable being used anywhere in `tensorflowpredict.cpp`

src/algorithms/machinelearning/tensorflowpredictmusicnn.h
```diff
@@ -53,7 +53,8 @@ class TensorflowPredictMusiCNN : public AlgorithmComposite {
   ~TensorflowPredictMusiCNN();
 
   void declareParameters() {
-    declareParameter("graphFilename", "the name of the file containing the model to use", "", Parameter::STRING);
+    declareParameter("graphFilename", "the name of the file from which to load the TensorFlow graph", "", "");
```
dbogdanov: Any particular reason for `""`? In many other algorithms we are using Parameter::STRING:

```
./src/algorithms/highlevel/gaiatransform.h:    declareParameter("history", "gaia2 history filename", "", Parameter::STRING);
./src/algorithms/io/monowriter.h:    declareParameter("filename", "the name of the encoded file", "", Parameter::STRING);
./src/algorithms/io/monowriter.h:    declareParameter("filename", "the name of the encoded file", "", Parameter::STRING);
./src/algorithms/io/yamlinput.h:    declareParameter("filename", "Input filename", "", Parameter::STRING);
./src/algorithms/io/audioloader.h:    declareParameter("filename", "the name of the file from which to read", "", Parameter::STRING);
./src/algorithms/io/audioloader.h:    declareParameter("filename", "the name of the file from which to read", "", Parameter::STRING);
./src/algorithms/io/monoloader.h:    declareParameter("filename", "the name of the file from which to read", "", Parameter::STRING);
./src/algorithms/io/monoloader.h:    declareParameter("filename", "the name of the file from which to read", "", Parameter::STRING);
./src/algorithms/io/metadatareader.h:    declareParameter("filename", "the name of the file from which to read the tags", "", Parameter::STRING);
./src/algorithms/io/audiowriter.h:    declareParameter("filename", "the name of the encoded file", "", Parameter::STRING);
./src/algorithms/io/audiowriter.h:    declareParameter("filename", "the name of the encoded file", "", Parameter::STRING);
./src/algorithms/io/eqloudloader.h:    declareParameter("filename", "the name of the file from which to read", "", Parameter::STRING);
./src/algorithms/io/eqloudloader.h:    declareParameter("filename", "the name of the file from which to read", "", Parameter::STRING);
./src/algorithms/io/easyloader.h:    declareParameter("filename", "the name of the file from which to read", "", Parameter::STRING);
./src/algorithms/io/easyloader.h:    declareParameter("filename", "the name of the file from which to read", "", Parameter::STRING);
./src/algorithms/extractor/freesoundextractor.h:    declareParameter("profile", "profile filename. If specified, default parameter values are overwritten by values in the profile yaml file. If not specified (empty string), use values configured by user like in other normal algorithms", "", Parameter::STRING);
./src/algorithms/extractor/musicextractor.h:    declareParameter("profile", "profile filename. If specified, default parameter values are overwritten by values in the profile yaml file. If not specified (empty string), use values configured by user like in other normal algorithms", "", Parameter::STRING);
```

src/algorithms/tonal/pitchcrepe.cpp
```diff
@@ -17,16 +17,16 @@
...
 using namespace standard;
 
-const char* CREPE::name = "CREPE";
-const char* CREPE::category = "Pitch";
-const char* CREPE::description = DOC(
-  "This algorithm estimates pitch using CREPE models.\n"
+const char* PitchCREPE::name = "PitchCREPE";
+const char* PitchCREPE::category = "Pitch";
+const char* PitchCREPE::description = DOC(
+  "This algorithm estimates pitch from monophonic audio signals using CREPE models.\n"
```
dbogdanov, line 29: "from" -> "of" or "for"?

test/src/unittests/tonal/test_pitchcrepe.py
```diff
@@ -47,23 +47,41 @@ def testRegression(self):
...
+                                                 'output': 'model/classifier/Sigmoid',
+                                                })  # input do not exist in the model
+        self.assertConfigureFails(PitchCREPE(), {'graphFilename': 'wrong_model_name',
+                                                 'input': 'frames',
+                                                 'output': 'model/classifier/Sigmoid',
+                                                })  # the model does not exist
+
+    def testEmptySignal(self):
+        model = join(testdata.models_dir, 'crepe', 'crepe-tiny-1.pb')
+        self.assertComputeFails(PitchCREPE(graphFilename=model), [])
```
dbogdanov, line 68: Is it conceptually fine for us that it fails on empty input? Is this consistent with other pitch estimation algorithms such as PitchYinProbabilistic? What is the minimum length required not to fail and is this information easily available to a user?

dbogdanov (changes requested):
Left some more comments. 

Btw, I see some previous changes not related to CREPE in this PR.

src/algorithms/machinelearning/tensorflowpredictmusicnn.h
palonso: They do different things:
-  `Parameter::STRING` means that by default the variable is not initialized
- `""` means that by default the variable is initialized to an empty string

In #1161 we modified the logic of `TensorflowPredict` so that the configuration is skipped when the parameters  `graphFilename` and `savedModel` are `""`, mainly  facilitate reusing the logic in `TensorflowPredict` instead of reimplementing it in dependent algorithms (e.g., `PitchCrepe`, `TensorflowPredictMusiCNN`...)

test/src/unittests/tonal/test_pitchcrepe.py
palonso: Failing in empty input is more consistent with other pitch algorithms:
```
test/src/unittests/sfx/test_pitchsalience.py:
  133  
  134:     def testEmpty(self):
  135          self.assertComputeFails(PitchSalience(), **[])**

test/src/unittests/tonal/test_pitchyin.py:
  26  
  27:     def testEmpty(self):
  28          self.assertComputeFails(PitchYin(), [])

test/src/unittests/tonal/test_pitchyinfft.py:
  26  
  27:     def testEmpty(self):
  28          self.assertComputeFails(PitchYinFFT(), [])

test/src/unittests/tonal/test_pitchyinprobabilistic.py:
  25  
  26:     def testEmpty(self):
  27          self.assertComputeFails(PitchYinProbabilistic(), [])
```

doc/sphinxdoc/research_papers.md
palonso: Maybe it was a local issue? This PR doesn't modify that file and below I can see `This branch has no conflicts with the base branch`

src/algorithms/machinelearning/tensorflowpredict.h
palonso: Addressed in #1186

PR #1196, opened 2021-10-28
adding new tests to pitch melody

cvf-bcn-gituser:
Adding new draft tests (Work in progress)


cvf-bcn-gituser:
@dbogdanov , this is in draft mode pending review of pitch contour segmentation, which has similar unit tests.
Some improvements there might also need to be applied here.

PR #1214, opened 2021-12-30
Handle unsuccessful load of savedModel

palonso:
Throw an exception if a `savedModel` is not loaded successfully.
This solution mimics the way we handle unsuccessful load of protocol buffers (`graphFilename`)

### DIFF
--- a/src/algorithms/machinelearning/tensorflowpredict.cpp
+++ b/src/algorithms/machinelearning/tensorflowpredict.cpp
@@ -129,6 +129,10 @@ void TensorflowPredict::openGraph() {
       _savedModel.c_str(), &tags_c[0], (int)tags_c.size(),
       _graph, NULL, _status);
 
+      if (TF_GetCode(_status) != TF_OK) {
+        throw EssentiaException("TensorflowPredict: Error importing SavedModel. ", TF_Message(_status));
+      }
+
     E_INFO("Successfully loaded SavedModel: `" << _savedModel << "`");
     return;
   }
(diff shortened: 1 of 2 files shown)


palonso:
@dbogdanov, two questions:
Tensorflow [calls](https://www.tensorflow.org/guide/saved_model) the format `SavedModel`. Our parameter is called `savedModel` to follow the convention of starting with lowercase. I'm using `SavedModel` to refer to the format and `savedModel` for the parameter. Is this misleading?

The derivate algorithms (e.g., `TensorflowPredictMusiCNN`) have similar `testInvalidParam` that I haven't updated, as they just rely on `TensorflowPredict` and that's already tested. Do you think it makes sense to keep this test redundancy, or should I just remove the additional tests in a separate PR?

dbogdanov:
In such cases, we could use dummy tests (e.g., `testDummyInvalidParam`) covering redundant tests, with a comment about algorithm dependencies and where the test is covered in the source code for the test.

src/algorithms/machinelearning/tensorflowpredict.cpp
```diff
@@ -129,6 +129,10 @@ void TensorflowPredict::openGraph() {
       _savedModel.c_str(), &tags_c[0], (int)tags_c.size(),
       _graph, NULL, _status);
 
+      if (TF_GetCode(_status) != TF_OK) {
+        throw EssentiaException("TensorflowPredict: Error importing SavedModel. ", TF_Message(_status));
```
dbogdanov, line 133: Following your comment, maybe we can explicitly say "...SavedModel specified in the `_savedModel` parameter." to avoid any confusion.

dbogdanov (changes requested):

PR #1216, opened 2021-12-30
Support SavedModels in Crepe

palonso:
This change makes this algorithm consistent with other algorithms from the TensorflowPredict* family


palonso:
The proposed changes make the Crepe algorithms consistent with #1161

dbogdanov:
This updates also adds `sicnn/VGG91_a125bis.pb` model file. Is it relevant for this PR?

palonso:
No, that model shouldn't be here. I'll open a clean PR without it (#1231)

PR #1220, opened 2022-01-03
Implement parameters SaverFilename and SaverFilenameSet

palonso:
This PR is a follow-up to #1161 that extends the support of TensorFlow `SavedModel` format.
While #1161 allowed reading `SavedModel`s, we found that sometimes those models require a `saver_filename` input tensor.
We found that this tensor is required when the `SavedModel` is created via:
- TensorFlow 2.X/Keras APIs (https://www.tensorflow.org/tutorials/keras/save_and_load)
- ONNX-TF backend (https://github.com/onnx/onnx-tensorflow)

However, we found that TensorFlow 1.X offers the possibility to transform a frozen model (.pb) to `SavedModel` and in this case `saver_filename` is not required.

According to [this blog](https://towardsdatascience.com/include-training-operations-in-saved-models-with-tensorflow-2-6494d304036d), the goal of `saver_filename` is to provide `SavelModel`s with a capability to be saved from low level APIs, but as it is not documented this feature would be probably changed or removed in the future.

For now, we are not interested in having model saving capabilities (as we don't support any way to modify the weights), so we will just pass the model a `saver_filename` input tensor string when needed to be able to do inference.
The value of this string can be modified with  the `saverFilename` parameter, but this won't have any effect for now, and it is only there considering the case in which we want to support model finetuning/saving in the future.
To support all the known scenarios, we provide a boolean parameter `saverFilenameSet` to  prevent passing `saver_filename` to the model when needed (e.g., in the `SavedModel`s we created from our .pbs for testing purposes).

### DIFF
--- a/src/algorithms/machinelearning/tensorflowpredict.cpp
+++ b/src/algorithms/machinelearning/tensorflowpredict.cpp
@@ -61,6 +61,8 @@ void TensorflowPredict::configure() {
   _isTraining = parameter("isTraining").toBool();
   _isTrainingName = parameter("isTrainingName").toString();
   _squeeze = parameter("squeeze").toBool();
+  _saverFilenameSet = parameter("saverFilenameSet").toBool();
+  _saverFilename = parameter("saverFilename").toString();
 
   (_isTrainingName == "") ? _isTrainingSet = false : _isTrainingSet = true;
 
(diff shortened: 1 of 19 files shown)


palonso:
I think I found a way to perform inference with `savedModels` without needing a `saver_filename` input. I'll close this PR while I dig further into this.

PR #1212, opened 2021-12-28
New doc page for Essentia TensorFlow models

dbogdanov:
New page with a complete overview of all available models.

PR #1222, opened 2022-01-05
Docs tensorflow models update

palonso:
Fixed issues discussed in #1212
- [x] Fix minor text problems
- [x] Improve style of the download link
- [x] Add index for the different tasks
- [x] Refactor name conventions

TODO:
- [ ]  add small snippets with the code to use each model


palonso:
@dbogdanov `Spleeter` and `OpenL3` doesn't have dedicated algorithms yet. Should we add the code to do inference with them (~150 lines for OpeneL3) or wait until we have those algorithms?

PR #1234, opened 2022-02-11
Add patchSize parameter

palonso:
Adds a  `patchSize` parameter so that `TensorflowPredictMusiCNN` and TensorflowPredictVGGish` can be use with derivate models expecting different input patch sizes.

Right now, we test this functionality  by assessing that an incorrect batch size produces an exception in VGGish. This approach can't be followed with MusiCNN as it accepts a dynamic input patch size.


src/algorithms/machinelearning/tensorflowpredictmusicnn.h
```diff
@@ -113,6 +114,7 @@ class TensorflowPredictMusiCNN : public Algorithm {
     declareParameter("patchHopSize", "number of frames between the beginnings of adjacent patches. 0 to avoid overlap", "[0,inf)", 93);
     declareParameter("lastPatchMode", "what to do with the last frames: `repeat` them to fill the last patch or `discard` them", "{discard,repeat}", "discard");
     declareParameter("accumulate", "when true it runs a single Tensorflow session at the end of the stream. Otherwise a session is run for every new patch", "{true,false}", false);
+    declareParameter("patchSize", "number of frames to feed to the model on each call. This parameter should match the model's expected input shape.", "[0,inf)", 187);
```
dbogdanov, line 117: Same here

dbogdanov (approved):

PR #1236, opened 2022-02-17
TensorflowPredictEffnetDiscogs

palonso:
A wrapper to `TensorflowPredict` specific for the  `EffnetDiscogs` family of models.

This algorithm features:
- Default input/output names specific to the `EffnetDiscogs` model.
- A function to zero-pad the input signal so that we can operate with a fixed batch size without losing the last frames


src/algorithms/machinelearning/tensorflowpredicteffnetdiscogs.cpp
```diff
@@ -0,0 +1,282 @@
...
+  "This algorithm makes predictions using EffnetDiscogs-based models.\n"
+  "\n"
+  "Internally, it uses TensorflowInputMusiCNN for the input feature extraction "
+  "(mel-spectrograms). It feeds the model with patches of 128 frames and "
+  "jumps a constant amount of frames determined by `patchHopSize`.\n"
+  "\n"
+  "By setting the `batchSize` parameter to -1 the patches are stored to run a single "
+  "TensorFlow session at the end of the stream. This allows to take advantage "
+  "of parallelization when GPUs are available, but at the same time it can be "
+  "memory exhausting for long files. This option is not be supported by some EffnetDiscogs models.\n"
```
dbogdanov, line 144: "is not supported by some"

test/src/unittests/machinelearning/test_tensorflowpredicteffnetdiscogs.py
```diff
@@ -0,0 +1,80 @@
...
+        for last_batch_mode in ("discard", "zeros", "same"):
+            model = TensorflowPredictEffnetDiscogs(
+                graphFilename=join(testdata.models_dir, 'effnetdiscogs', 'effnetdiscogs-bs64-1.pb'),
+                lastBatchMode=last_batch_mode,
+                batchSize=batch_size,
+            )
+
+            # It'd be nice to test other durations but we are limited by inference times,
+            # especially in CPU.
+            seconds = 150
```
dbogdanov, line 63: If this test is still slow,  can we use smaller batch_size and therefore smaller input audio size?

dbogdanov (changes requested):

test/src/unittests/machinelearning/test_tensorflowpredicteffnetdiscogs.py
palonso: The EffnetDiscogs that we are testing here only supports batch_size=64.

src/algorithms/machinelearning/tensorflowpredicteffnetdiscogs.cpp
```diff
@@ -0,0 +1,284 @@
...
+  if (_lastPatchMode == "repeat") {
+    nPatches = 1 + ceil(patchHops);
+  } else if (_lastPatchMode == "discard") {
+    nPatches = 1 + floor(patchHops);
+  } else {
+    throw EssentiaException("TensorflowPredictEffnetDiscogs: incorrect `lastPatchMode`");
+  }
+
+  // Patches to batches.
+  int nBatches = ceil((Real)nPatches / (Real)_batchSize);
```
dbogdanov, line 262: `_batchSize` = 0 is accepted by the parameter definition `[-1,inf)`, so this would lead to division by zero.

test/src/unittests/machinelearning/test_tensorflowpredicteffnetdiscogs.py
```diff
@@ -0,0 +1,80 @@
...
+# details.
+#
+# You should have received a copy of the Affero GNU General Public License
+# version 3 along with this program. If not, see http://www.gnu.org/licenses/
+
+
+from essentia_test import *
+
+
+class TestTensorFlowPredictEffnetDiscogs(TestCase):
```
dbogdanov, line 24: Any test for what happens in the case of batch size mismatch?

palonso: TF throws a shape mismatch exception in that case.
I don't think this should be tested because it is the responsibility of the user to know if the model requires a specific batch size or not. But we should clearly mention this in the models' page.

I designed the algorithm to be batch-size agnostic because in the future we may be interested in creating new versions of the model with different batch sizes. I even allowed the `accumulate mode` (the batch size is dynamically set to fit the whole input stream) to be consistent with other algorithms and because we may find the way to support dynamic dimensions in the future.

dbogdanov:
@palonso The `batchSize` parameter ranges is somewhat misleading. We allow `-1` as a special case, but `0` is prohibited. Why not using `batchSize=0` for that instead of `-1`?

palonso:
I didn't want to use 0 because it is a valid dimension shape in Numpy or Tensorflow.
NumPy functions use `-1` to reflect an unknown dimension (e.g., [reshape](https://numpy.org/doc/stable/reference/generated/numpy.reshape.html)).
TensorFlow uses `None` for this, but using an int is more convenient in C++.
Also, I find that  `-1` conveys a bit better the idea that we are not givin an exact value but something else.

dbogdanov:
I see this is consistent with PitchCrepe (TensorflowPredictCrepe) and TempoCNN (TensorflowPredictTempoCNN) which is nice. What i dislike is that it breaks the semantics of the `[-1,inf)` which should indicate an expected valid range and that `batchSize=0` already reads like a special case given the current description of the parameter. And we never use `-1` for anything like this in other algorithms.


One option is to mention that the exception will be thrown for `batchSize=0` in the DOC strings in all such algorithms. Another option is allow both 0 and -1 with the same meaning.

Here is few completely unrelated random examples of what batch size 0 [can](https://developer.apple.com/documentation/coredata/nsfetchrequest/1506558-fetchbatchsize) [mean](https://docs.oracle.com/cd/E85694_01/ODPNT/BulkCopyBatchSize.htm) for different users from where we can see that using `0` could also be common sense for users not directly involved with TensorFlow.

palonso:
Both options make sense for me but maybe allowing both values to mean the same thing makes more sense from the perspective of the valid range.
If you prefer this option I would merge this PR and open a new one with the changes to support the new functionality with consistency across the following algorithms:
`TensorflowPredictMusiCNN`, `TensorflowPredictVGGish`, `TensorflowPredictCREPE`, `TensorflowPredictTempoCNN`, `PitchCREPE`, `TempoCNN`, and `VectorRealToTensor`

dbogdanov:
This sounds good, let's use both values and open a separate PR. Merging this.

PR #1252, opened 2022-04-27
KeyExtractor tests and Key algorithm fixes.

seeker-Liu:


seeker-Liu:
Another possible fix is to throw a specific error when the input is empty. However I am not sure where and how to add it so and left it out.

src/algorithms/tonal/key.cpp
```diff
@@ -367,21 +367,21 @@ void Key::compute() {
...
     max = maxMinor;
     max2 = max2Minor;
     }
 
 	else if (maxOther > maxMajor && maxOther > maxMinor) {
-    keyIndex = (int) (keyIndexOther * 12 / pcpsize + 0.5);
+    keyIndex = std::roundl(keyIndexOther * 12 / pcpsize);
     scale = MAJMIN;
     max = maxOther;
     max2 = max2Other;
```
dbogdanov, line 387: We commonly use `round` in other places. Why using `roundl` instead?

dbogdanov (changes requested):

dbogdanov:
> Another possible fix is to throw a specific error when the input is empty. However I am not sure where and how to add it so and left it out.

Some other algorithms (`TonalExtractor`, `RhythmExtractor2013`, etc.) would similarly fail with exception on empty input. We'll need to standardize the behavior.

src/algorithms/tonal/key.cpp
seeker-Liu: I was thinking about std::lround...
Will correct this.

src/algorithms/extractor/keyextractor.cpp
```diff
@@ -207,14 +207,25 @@ void KeyExtractor::createInnerNetwork() {
...
+  if(audio.size() == 0)
+  {
+    /* It is only a temporary solution. After the empty pool issue (https://github.com/MTG/essentia/pull/1252#issuecomment-1115062252)
+     * got resolved, this if-clause should be safe be removed.
+     */
+    key = "C";
+    scale = "major";
+    strength = 0.0;
+    return;
+  }
```
dbogdanov, line 227: Let's just throw an exception like other algorithms (TonalExtractor, RhythmExtractor2013) do.

src/algorithms/tonal/key.cpp
```diff
@@ -367,21 +367,21 @@ void Key::compute() {
...
     scale = MINOR;
     max = maxMinor;
     max2 = max2Minor;
     }
 
 	else if (maxOther > maxMajor && maxOther > maxMinor) {
-    keyIndex = std::roundl(keyIndexOther * 12 / pcpsize);
+    keyIndex = std::lround(keyIndexOther * 12 / pcpsize);
     scale = MAJMIN;
     max = maxOther;
```
dbogdanov, line 386: `keyIndex` is an int, not `long int`, therefore use `round` instead of `lround`. This will be consistent with code in other algorithms doing similar rounding.

dbogdanov (changes requested):

src/algorithms/extractor/keyextractor.cpp
seeker-Liu: This is already modified. You were checking the older version.

src/algorithms/tonal/key.cpp
seeker-Liu: By using std::round we are relying on implicit conversion (float to integer), and this conversion, when the value is out of the range of int type, is a UB. Using std::lround() prevent that UB because std::lround with set the floating environemtn error if the value is too big to be represented by long int, and convert a long int with too big value to int is only unspecified (pre C++20) or well-defined (since C++20). 
Anyway the difference is very insignificant. When the value is within the range of int both ways give correct result, and I believe using std::lround is safer.

dbogdanov: In practice we'll never get too big values in this case. In general I agree which such changes, but they should be systematically applied to the entire code base.

PR #1267, opened 2022-07-20
ChordsDetectionBeats tests

seeker-Liu:


test/src/unittests/tonal/test_chordsdetectionbeats.py
```diff
@@ -0,0 +1,81 @@
...
+# version 3 along with this program. If not, see http://www.gnu.org/licenses/
+
+from essentia_test import *
+import numpy as np
+
+
+class TestChordsDetectionBeats(TestCase):
+    def testEmpty(self):
+        with self.assertRaises(EssentiaException):
+            ChordsDetectionBeats()(np.zeros(shape=(0, 12), dtype=np.single), np.zeros(shape=(0,), dtype=np.single))
```
dbogdanov, line 25: There are several options here we can check:
- empty input `pcp` and `ticks` (tested now)
- empty input `pcp` but non-empty `ticks`
- empty input `ticks` but non-empty `pcp`

Also, from the DOC description of the algorithm it is not clear what are the requirements on the input `ticks`. Let's add this information to the DOC string. What happens if there is a single `ticks` value or no values?

src/algorithms/tonal/chordsdetection.cpp
```diff
@@ -27,7 +27,8 @@ namespace standard {
 
 const char* ChordsDetection::name = "ChordsDetection";
 const char* ChordsDetection::category = "Tonal";
-const char* ChordsDetection::description = DOC("This algorithm estimates chords given an input sequence of harmonic pitch class profiles (HPCPs). It finds the best matching major or minor triad and outputs the result as a string (e.g. A#, Bm, G#m, C). This algorithm uses the Sharp versions of each Flatted note (i.e. Bb -> A#).\n"
+const char* ChordsDetection::description = DOC("This algorithm estimates chords given an input sequence of harmonic pitch class profiles (HPCPs). It finds the best matching major or minor triad and outputs the result as a string (e.g. A#, Bm, G#m, C). The note names used as output is listed follow:\n"
```
dbogdanov, line 30: Some changes: 

- we need to update this example strings too: "(e.g. A#, Bm, G#m, C)" --> "(e.g. Bb, Bm, Abm, C)" 

- "The note names used as output is listed follow: ... " --> `"The following note names are used in the output: \"A\", \"Bb\", \"B\", \"C\", \"C#\", \"D\", \"Eb\", \"E\", \"F\", \"F#\", \"G\", \"Ab\"."`

dbogdanov (changes requested):

PR #1292, opened 2022-11-24
Prefix for vamp linux install path

miker2049:
This change makes sure the user can define the location of the vamp 3rd-party plugin at configure-time


dbogdanov:
Thanks!

PR #1306, opened 2023-02-17
Guard against segfaults in PitchYIN FFT

leighsmith:
Now guards against positions or amplitudes vectors being empty, raising an exception, to avoid segfaults.


src/algorithms/tonal/pitchyinfft.cpp
```diff
@@ -185,8 +185,15 @@ void PitchYinFFT::compute() {
     _peakDetect->output("amplitudes").set(_amplitudes);
     _peakDetect->compute();
     try {
```
dbogdanov, line 187: Thanks for the proposed modification. This code can be simplified, removing the enclosing try/catch and instead using the proposed check for vector sizes and throwing an exception.

The exception text "PitchYinFFT: it appears that no peaks were found by PeakDetection. If you read this message, PLEASE, report this issue to the developers with an example of audio on which it happened." is more informative, so we should use that.

No need to set ` tau = 0.` and `yinMin = 0.` before throwing the exception.

dbogdanov (changes requested):

dbogdanov:
Thank you @leighsmith!

PR #1307, opened 2023-02-17
Fix sample rate used for LPC computation.

leighsmith:
- Corrected LPC when run at non-default sample rate by passing on the sample rate of LPC to WarpedAutoCorrelation.


leighsmith:
@dbogdanov Can you please assign a reviewer to this very small patch? Many thanks!

dbogdanov (approved):

PR #1309, opened 2023-02-27
Update FSD-SINet model names

palonso:


doc/sphinxdoc/models.rst
```diff
@@ -163,7 +163,7 @@ Models:
     .. literalinclude:: ../../src/examples/python/models/scripts/audio-event-recognition/yamnet/audioset-yamnet-1_embeddings.py
 
 
-FSD50K
+FSD-SINet
```
dbogdanov, line 166: I guess we already discussed that before, but I liked that the FSD50K header was more explicit about the dataset version used and in most of other headers we specify the dataset used. Maybe we can combine both in this header (FSD50K-FSDSINet)? Same applies to models files names. If we assume that for our purposes `fsd` will always relate to the same taxonomy as used in the FSD50K version then the model naming is fine.

palonso: I think that FSD-SINet is actually more explicit since it referes to a data-model combination. Note that we explicitly mention that the FSD-SINet family of models rely on FSD50K in [this](https://github.com/MTG/essentia/blame/27bd293eeeeb3e8f555a759ebb11458d489233fb/doc/sphinxdoc/models.rst#L169) and [this](https://github.com/MTG/essentia/blame/27bd293eeeeb3e8f555a759ebb11458d489233fb/doc/sphinxdoc/models.rst#L214) lines. So I think naming the subsection FSD50K-FSDSINet is redundant.

The advantage of this change is that we get a hyperlink specific to model family. Otherwise we have a generic hyperlink for FSD50K models and we will have to keep expanding the naming convention section every time we get a new family of models.

I don't think we can assume that `fsd` refers to FSD50K only. A new model family `FSD-<new_architecture>` could be trained in a different subset of Freesound. But I think this is OK as far as we mention the specific dataset name as [here](https://github.com/MTG/essentia/blame/27bd293eeeeb3e8f555a759ebb11458d489233fb/doc/sphinxdoc/models.rst#L214) and in the models' metadata. We mention this issue with @edufonseca and @ffont, and we agreed to prioritize readability (FSD-SINet) over being explicit (FSD50K-SINet) in the model name.

dbogdanov: OK great, it is clear to me now!

Issue #1333, opened 2023-05-16
TensorflowPredictVGGish


seunggookim:
I would like to find corresponding patches for shallow layer activation (MaxPool). But, because they are flattened, it is difficult for me. For example, if I feed a short audio segment for a single patch (15600 samples at 16 kHz), `model/vggish/pool4/MaxPool` is (24, 512). But for a longer audio file, it is (24*#patches, 512). For me, it's unclear whether the first dimension, which is flattened, is along patches first ([p1f1, p2f1, p3f1, ..., p1f2, p2f2, p3f2, ...]) or features ([p1f1, p1f2, p1f3, ..., p2f1, p2f2, p2f3, ...]).

To figure this out myself, I manually segmented the audio file into 15600 samples before feeding them to the models as follows:
```
patchSize_smp = 15600      # 96 frames (95 frame hops + 1 frame)
patchHopSize_smp = 14880   # 93 frames (93 frame hops)
frameSize_smp = 400
frameHopSize_smp = 160
audio = MonoLoader(filename=audio_path, sampleRate=16000)()
audioSize_smp = len(audio)
startPoints_smp = np.arange(1, audioSize_smp - patchSize_smp, patchHopSize_smp) 
endPoints_smp = np.arange(patchSize_smp, audioSize_smp, patchHopSize_smp)

# extract activation
model = TensorflowPredictVGGish(graphFilename=model_path, output="model/vggish/embeddings")
activation = []
for start_smp, end_smp in zip(startPoints_smp, endPoints_smp):
    activation.append(model(audio[start_smp:end_smp+1]).copy())
activation = np.array(activation)

```
Because the final layer "embeddings" is clear to me, I compared the results for the final layer. What I tried got me the same number of patches, but slightly different activation values (r = 0.6349 to 1; mean r = 0.9743), which is still disturbing. Should I add anything to make these identical to the original function?

seunggookim:
Thanks @palonso!
I saw the `mel_features.py` in the [repo](https://github.com/DTaoo/VGGish/blob/f6a6ecf0891c915aa6ab8fdb20e08eb29c056b74/mel_features.py#L209) you linked and [another one](https://github.com/tensorflow/models/blob/35ed75e315a582640c5c25f870f4e23e9c35b8b9/research/audioset/vggish/mel_features.py#L209) (perhaps an official version?). 

[...]

PR #1338, opened 2023-05-26
Improve intermediate layer extraction explanation

palonso:
[TensorToVectorReal](https://essentia.upf.edu/reference/streaming_TensorToVectorReal.html) converts tensors to 2D arrays by flattening all axis but the last one into the first dimension.
model-specific prediction algorithms (e.g., [TensorflowPredictVGGish](https://essentia.upf.edu/reference/std_TensorflowPredictVGGish.html)) use this algorithm to return 2D arrays since they are primarily intended for time-wise predictions or embeddings. However, it is possible to use these algorithms to extract intermediate layers of the models that may have more than two dimensions. In this case, all dimensions but the last one will be flattened. To address this:
- TensorToVectorReal throws a warning in case it flattens a dimension.
- We added notes explaining this behavior to the algorithms potentially affected.

Note that it is also possible to retrieve intermediate layers with their original shape using [TensorflowPredict](https://essentia.upf.edu/reference/std_TensorflowPredict.html) as discussed [here](https://github.com/MTG/essentia/issues/1333#issuecomment-1563932713).


src/algorithms/machinelearning/tensorflowpredicteffnetdiscogs.cpp
```diff
@@ -156,6 +156,11 @@ const char* TensorflowPredictEffnetDiscogs::description = DOC(
   "Note: This algorithm does not make any check on the input model so it is "
   "the user's responsibility to make sure it is a valid one.\n"
   "\n"
+  "Note: The output of this algorithm is 2D, which is suitable for extracting embeddings or "
+  "class activations (the output shape is, e.g., [time, number of classes]). If the output "
+  "parameter is set to an intermediate layer with more dimensions, the output will be "
+  "flattened to 2D.\n"
```
dbogdanov, line 162: Rephrased version (trying to simplify):

> Note: The algorithm outputs a time series of class activations or embedding vectors, with a 2D shape [time, feature vector]. Feature vector values will be flattened if the `output` parameter is set to extract an intermediate layer with multiple dimensions.

src/algorithms/machinelearning/tensorflowpredictfsdsinet.cpp
```diff
@@ -159,12 +159,17 @@ const char* TensorflowPredictFSDSINet::description = DOC(
   "Note: This algorithm does not make any check on the input model so it is "
   "the user's responsibility to make sure it is a valid one.\n"
   "\n"
+  "Note: The output of this algorithm is 2D, which is suitable for extracting embeddings or "
+  "class activations (the output shape is, e.g., [time, number of classes]). If the output "
+  "parameter is set to an intermediate layer with more dimensions, the output will be "
+  "flattened to 2D.\n"
+  "\n"
```
dbogdanov, line 166: Same  comments as for TensorflowPredictEffnetDiscogs

src/algorithms/machinelearning/tensorflowpredictmusicnn.cpp
```diff
@@ -158,6 +158,11 @@ const char* TensorflowPredictMusiCNN::description = DOC(
   "Note: This algorithm does not make any check on the input model so it is "
   "the user's responsibility to make sure it is a valid one.\n"
   "\n"
+  "Note: The output of this algorithm is 2D, which is suitable for extracting embeddings or "
+  "class activations (the output shape is, e.g., [time, number of classes]). If the output "
+  "parameter is set to an intermediate layer with more dimensions, the output will be "
+  "flattened to 2D.\n"
```
dbogdanov, line 164: Same comment as for TensorflowPredictEffnetDiscogs

src/algorithms/machinelearning/tensorflowpredictvggish.cpp
```diff
@@ -156,6 +156,11 @@ const char* TensorflowPredictVGGish::description = DOC(
   "Note: This algorithm does not make any check on the input model so it is "
   "the user's responsibility to make sure it is a valid one.\n"
   "\n"
+  "Note: The output of this algorithm is 2D, which is suitable for extracting embeddings or "
+  "class activations (the output shape is, e.g., [time, number of classes]). If the output "
+  "parameter is set to an intermediate layer with more dimensions, the output will be "
+  "flattened to 2D.\n"
```
dbogdanov, line 162: Same comment as for TensorflowPredictEffnetDiscogs

src/algorithms/standard/tensortovectorreal.cpp
```diff
@@ -66,6 +68,11 @@ AlgorithmStatus TensorToVectorReal::process() {
     _timeStamps = tensor.dimension(2);
     _featsSize = tensor.dimension(3);
 
+    if (_channels != 1 && !_warned) {
+        E_WARNING("TensorToVectorReal: The channel axis (dimension 1) of the input tensor has size larger than 1, but the output of this algorithm is 2D. The batch, channel, and time axes (dimensions 0, 1, 2) will be flattened to the first dimension of the output matrix.");
```
dbogdanov, line 72: We output a vector of vector of reals, so the "matrix" terminology may be misleading.

dbogdanov (changes requested):
This looks good! I've left a proposal to improve the description of the algorithms' output in the DOC string.

PR #1339, opened 2023-06-01
window type is changed to hamming in discontinuity detection.

AhmetOguzOzturk:
In DiscontinuityDetector.cpp file, triangular window type is changed to Hamming, after the study of my master thesis.

### DIFF
--- a/src/algorithms/audioproblems/discontinuitydetector.cpp
+++ b/src/algorithms/audioproblems/discontinuitydetector.cpp
@@ -45,7 +45,7 @@ void DiscontinuityDetector::configure() {
   _medianFilter->configure(INHERIT("kernelSize"));
   _LPC->configure(INHERIT("order"));
   _windowing->configure("size", _frameSize, "zeroPhase", false, "type",
-                        "triangular");
+                        "hamming");
 
   if (_frameSize <= _order)
     throw(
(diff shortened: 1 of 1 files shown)


palonso:
Thank you for your contribution @AhmetOguzOzturk,

I tested the new behavior and found that [testNoOverlap](https://github.com/MTG/essentia/blob/37c97adbf53b24e1f1e137d7b1f7fa8999efbf2c/test/src/unittests/audioproblems/test_discontinuitydetector.py#L94)  is not passing with the new window type. This means that the algorithm behaves worse under no-overlap conditions with the new window type. Since I can't think of a scenario where a non-overlapped analysis is particularly useful I propose the following changes:

- Explicitly recommend using half-overlapped frames in the [algorithm documentation](https://github.com/MTG/essentia/blob/37c97adbf53b24e1f1e137d7b1f7fa8999efbf2c/src/algorithms/audioproblems/discontinuitydetector.cpp#L29). We can add the following lines:

> 
> The recommended pipeline is as follows:
> 
> ```
> MonoLoader(sampleRate=16000, filename="audio.wav") >> FrameCutter(frameSize=512, hopSize=256, startFromZero=True) >> DiscontinuityDetector(frameSize=512, hopSize=256)
> 
> ```

 

- Remove the [testNoOverlap](https://github.com/MTG/essentia/blob/master/test/src/unittests/audioproblems/test_discontinuitydetector.py#L94), since we no longer recommend this type of analysis.

Any thoughts @dbogdanov ?

dbogdanov:
If the frames should be always half-overlapped, why do we need the [`hopSize`](https://github.com/MTG/essentia/blob/37c97adbf53b24e1f1e137d7b1f7fa8999efbf2c/src/algorithms/audioproblems/discontinuitydetector.h#L70) parameter?

dbogdanov:
I agree with your suggestions. We can either remove or update the expected ground truth for `testNoOverlap`.

PR #1364, opened 2023-08-11
Fix Cross-Compile Errors + CVE-2022-37434 security patch

Ma5onic:
Version 1.2.13 has this key update:

Fix a bug when getting a gzip header extra field with `inflateGetHeader()`. This remedies [CVE-2022-37434.](https://nvd.nist.gov/vuln/detail/CVE-2022-37434)

> IMPORTANT: Due to the first bug fix, any installations of 1.2.12 or earlier should be replaced with 1.2.13.


Ma5onic:
Note: python bindings still don't compile

PR #1378, opened 2023-09-19
Framebuffer - Issue in PitchYINFFT when no peaks are detected.

xaviliz:
Small fix in PitchYINFFT.
Sorry, I reverted the first commit because I had an issue in my fork.


xaviliz:
Ok, this is the correct PR. Thanks!

PR #1382, opened 2023-10-20
Add audio2pitch and pitch2midi algorithms.

xaviliz:
- Audio2pitch: includes PitchYinFFT, RMS and FrameBuffer algorithms to provide an accurate pitch estimation for RT purpouses.
- Pitch2MIDI: just convert pitch and rms features to MIDI features (midi note, closest note name, closest pitch, difference between pitch and closest pitch in Hz, difference between pitch and closest pitch in cents and velocity).


dbogdanov:
Closing the PR as it is replaced with new ones by @xaviliz

Issue #1325, opened 2023-04-25
Support essentia install on macOS


dgoldenberg-audiomack:
There are a lot of tickets in this repo and the `homebrew-essentia` repo which report various issues with installing Essentia on macOS. I'd like to propose that this be done under a particular ticket like this one. This should work just like a *nix setup/install.

regorxxx:
There are thousands of tickets about being impossible to build essentia (+gaia) on multiple systems and zero support... still waiting for a single answer to any of them.

With all due respect, I don't see the point of having constant updates when the repository and can not be built by most people since years and all instructions are outdated.

dbogdanov:
Python wheels for Apple Silicon are now available on PyPI ([essentia](https://pypi.org/project/essentia/), [essentia-tensorflow](https://pypi.org/project/essentia-tensorflow/)).

Dazzid:
Certainly! I'm using a MacOS Ventura 13.0 M1 Chip. 
Python 3.10.6 running in a conda env. 

Some more info
`Package                       Version
----------------------------- -------------
absl-py                       1.3.0
aiohttp                       3.8.3
aiosignal                     1.3.1
anyio                         3.6.2
appdirs                       1.4.4
appnope                       0.1.3
argon2-cffi                   21.3.0
argon2-cffi-bindings          21.2.0
asttokens                     2.1.0
astunparse                    1.6.3
async-timeout                 4.0.2
attrs                         22.1.0
audioread                     3.0.0
backcall                      0.2.0
backports.functools-lru-cache 1.6.4
beautifulsoup4                4.11.1
bleach                        1.5.0
blinker                       1.5
blobfile                      2.0.0
bokeh                         2.4.3
brotlipy                      0.7.0
bz2file                       0.98
cached-property               1.5.2
cachetools                    5.2.0
certifi                       2022.12.7
cffi                          1.15.1
chardet                       5.0.0
charset-normalizer            2.1.1
click                         8.1.3
cloudpickle                   2.2.0
contourpy                     1.0.6
cryptography                  38.0.3
cycler                        0.11.0
datasets                      2.8.0
debugpy                       1.6.3
decorator                     5.1.1
defusedxml                    0.7.1
dill                          0.3.6
docker-pycreds                0.4.0
dopamine-rl                   3.0.1
entrypoints                   0.4
essentia                      2.1b6.dev1110
essentia-tensorflow           2.1b6.dev1110
etils                         0.8.0
executing                     1.2.0
[...]

PR #1383, opened 2023-10-24
Add changes to support building arm64 macos wheels

palonso:
In theory, Apple Silicon wheels can be cross-compiled from the intel mac machines (however the can't be tested):
https://cibuildwheel.readthedocs.io/en/stable/faq/#apple-silicon

Additionally, we replace `os.system` calls by `subprocess.run` so that we can notice when one of these subprocesses fail.

### DIFF
--- a/pyproject-tensorflow.toml
+++ b/pyproject-tensorflow.toml
@@ -44,7 +44,7 @@ test-command = "python -c 'import essentia; import essentia.standard; import ess
 [[tool.cibuildwheel.overrides]]
 select = "*macosx_arm64*"
 
-skip = ["pp*", "cp38-macosx_*"]
+skip = ["pp*"]
 
 environment = { PROJECT_NAME="essentia", ESSENTIA_PROJECT_NAME="${PROJECT_NAME}", ESSENTIA_WHEEL_SKIP_3RDPARTY=1, ESSENTIA_WHEEL_ONLY_PYTHON=1, ESSENTIA_MACOSX_ARM64=1 }
 
--- a/pyproject-tensorflow.toml
+++ b/pyproject-tensorflow.toml
@@ -4,6 +4,7 @@ manylinux-x86_64-image = "mtgupf/essentia-builds:manylinux2014_x86_64"
 
 # Only support x86_64 for essentia-tensorflow
 build = "cp**-manylinux_x86_64"
+# TODO: skipping Python 3.12 for now until we create manylinux wheels supporting this version.
 skip = ["pp*", "*-musllinux*", "*i686", "*cp312*"]
 
 environment = { PROJECT_NAME="essentia-tensorflow", ESSENTIA_PROJECT_NAME="${PROJECT_NAME}", ESSENTIA_WHEEL_SKIP_3RDPARTY=1, ESSENTIA_WHEEL_ONLY_PYTHON=1 }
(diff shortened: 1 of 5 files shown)


pyproject-tensorflow.toml
```diff
@@ -39,6 +39,19 @@ before-all = [
 test-command = "python -c 'import essentia; import essentia.standard; import essentia.streaming; from essentia.standard import MonoLoader, MetadataReader, YamlInput, Chromaprinter, TensorflowPredict'"
 
 
+[[tool.cibuildwheel.overrides]]
+select = "*arm64*"
```
dbogdanov: Make the override `select` more specific so that it only triggers macos builds (to avoid a conflict with future arm64 Linux builds).

dbogdanov (changes requested):

dbogdanov:
> In theory, Apple Silicon wheels can be cross-compiled from the intel mac machines (however the can't be tested).
I thinks this won't work because we rely on homebrew to install dependencies (and those would still be x86).

palonso:
> > In theory, Apple Silicon wheels can be cross-compiled from the intel mac machines (however the can't be tested).
> 
> I think this won't work because we rely on homebrew to install dependencies (and those would still be x86).

You are right. I was exploring [this hack](https://github.com/orgs/Homebrew/discussions/2843?sort=top#discussioncomment-2026915) to force installing arm64 libs for the cross-build case.

palonso:
After the release of the native arm64 MacOS worker ([macos-14](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/)), this PR has been updated to use it directly instead of to cross-compiling from an intel machine.

Main changes:
- Add `macos-14` worker for arm64 based (M*) MacOS wheels
- Bump the Intel MacOS builds `macos-11` [[deprecated](https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images)] -> `macos-12`
- Bump Actions checkout, and upload-artifact from `v3` -> `v4`, so they use node 20 instead of node 16 (that will reach the end of life [by Spring 2024](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/)).
- Bump cibuildwheels `11.4` -> `16.5` which internally upgrades `setup-python` to use node 20 instead of 16 and supports building wheels for Python 3.12.
- Manually copy libSLD2 dylibs into macos-14 wheels to prevent crashes (#1325).

dbogdanov:
There is an error in building the arm64 essentia-tensorflow wheels for MacOS. The `artifact-macos-14-pyproject-tensorflow.zip` contains files for main esssentia instead (e.g., `essentia-2.1b6.dev1161-cp310-cp310-macosx_11_0_arm64.whl`)

pyproject-tensorflow.toml
```diff
@@ -39,6 +42,40 @@ before-all = [
 test-command = "python -c 'import essentia; import essentia.standard; import essentia.streaming; from essentia.standard import MonoLoader, MetadataReader, YamlInput, Chromaprinter, TensorflowPredict'"
 
 
+[[tool.cibuildwheel.overrides]]
+select = "*macosx_arm64*"
+
+skip = ["pp*"]
+
+environment = { PROJECT_NAME="essentia", ESSENTIA_PROJECT_NAME="${PROJECT_NAME}", ESSENTIA_WHEEL_SKIP_3RDPARTY=1, ESSENTIA_WHEEL_ONLY_PYTHON=1, ESSENTIA_MACOSX_ARM64=1 }
```
dbogdanov, line 50: wrong project name, creates wrong wheel names.

dbogdanov (changes requested):

PR #1384, opened 2023-10-26
FrameBuffer algorithm

dbogdanov:
Adds implementation for the FrameBuffer algorithm.

@xaviliz please report if there are any issues/requests for the FrameBuffer behavior before we are merging this.


xaviliz:
Hi, it is working properly. NO ISSUES but we need to extend this class for the Midi Pool messages he have prototyped for NOTE_ON and NOTE_OFF definition.

PR #1400, opened 2024-02-06
PitchYinFFT updates (new weighting parameter, do not throw exception on empty peak detection)

xaviliz:
1) Add weighting parameter to select different weighting funcitons in PITCHYINFFT algorithm. The available choices are: 
    - the default weighting function
    - A-weighting
    - B-weighting
    - C-weighting
    - D-weighting
    - Z-weighting.

    Note that there is not experimental validation of better pitch estimation performance with new alternative weightings. 

2) Change behavior on empty spectrum peak detection: output zero pitch and confidence instead of throwing an exception.

@dbogdanov Edited for completeness.

### DIFF
--- a/src/algorithms/tonal/pitchyinfft.cpp
+++ b/src/algorithms/tonal/pitchyinfft.cpp
@@ -30,10 +30,38 @@ static const Real _freqsMask[] = {0., 20., 25., 31.5, 40., 50., 63., 80.,
 	1600., 2000., 2500., 3150., 4000., 5000., 6300., 8000., 9000., 10000.,
 	12500., 15000., 20000.,  25100};
 
-static const Real _weightMask[] = {-75.8, -70.1, -60.8, -52.1, -44.2, -37.5,
+static Real _weightMask[] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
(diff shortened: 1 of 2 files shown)


src/algorithms/tonal/pitchyinfft.cpp
```diff
@@ -189,8 +245,9 @@ void PitchYinFFT::compute() {
       yinMin = -_amplitudes[0];
     }
     else {
-      // TODO this should never happen, but some people reported it happening in their real time applications.
-      throw EssentiaException("PitchYinFFT: it appears that no peaks were found by PeakDetection. If you read this message, PLEASE, report this issue to the developers with an example of audio on which it happened.");
+      tau = 0.0; // it will provide zero-pitch and zero-pitch confidence.
```
dbogdanov, line 248: This change is out of scope of weighting functions.  At least would should add the description of this change to the PR.

xaviliz: Oh yes! this is something we need in the ODISEI project to avoid a crucial error. Indeed we would need to launch a negative pitch value. No problem I will redo what essentia does in this cases.

dbogdanov: This change had been already included in the merge.

PR #1403, opened 2024-02-27
Add Unit Tests for Tonal Extractor Algorithm

cvf-bcn-gituser:
The Extractor folder contains unit tests for key, level, music, rhythm descriptors and tuning frequency. Until now the tonal extractor has had no unit tests.

The unit tests added use the same general "boiler plate" of assertion tests that are typical of an extractor (silent test, valid config, etc.) as well as some tests on real audio, with some variation on config parameters. There is scope for adding more output checks on the more complex outputs such as chords_histogram, chords_progression. etc.

The idea behind capturing reference values  for "chord changes rate", "key strength"  for given audio files, is to support future regression testing on algorithm optimization and also to provide values to compare with other libraries.


cvf-bcn-gituser:
FYI: See above pull request @palonso , @dbogdanov

PR #1408, opened 2024-04-12
[feat] Support source code links from algorithm reference

dhunstack:
Add links to C++ source and header file for each algorithm
Screenshot 2024-04-12 at 12 17 22

### DIFF
--- a/doc/sphinxdoc/generate_reference.py
+++ b/doc/sphinxdoc/generate_reference.py
@@ -26,6 +26,7 @@ import os, re, subprocess
 import sys
 from pathlib import Path
 import nbformat
+import git
 
 
 std_algo_list = [ algo for algo in dir(essentia.standard) if algo[0].isupper() ]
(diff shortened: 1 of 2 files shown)


doc/sphinxdoc/generate_reference.py
```diff
@@ -160,6 +187,15 @@ def doc2rst(algo_doc, sphinxsearch=False):
         lines = [ algo_doc['name'], '=' * len(algo_doc['name']), '']
         lines += [algo_doc['mode'] + ' | ' + algo_doc['category'] + ' category', '']
 
+    links = source_links(algo_doc)
+    if links:
+        lines += ['Source Code',
```
dbogdanov, line 192: "Source code" - for consistency in capitalization in the documentation (e.g., see "See also" in the same algorithm page).

doc/sphinxdoc/generate_reference.py
```diff
@@ -160,6 +187,15 @@ def doc2rst(algo_doc, sphinxsearch=False):
         lines = [ algo_doc['name'], '=' * len(algo_doc['name']), '']
         lines += [algo_doc['mode'] + ' | ' + algo_doc['category'] + ' category', '']
 
+    links = source_links(algo_doc)
+    if links:
```
dbogdanov, line 191: I suggest moving this section to put it between the "Description" and "See also" sections because:
- source code makes more sense after reading basic algorithm description
- not everyone will want to inspect C++ source code

doc/sphinxdoc/generate_reference.py
```diff
@@ -147,6 +147,33 @@ def related_tutorials(algo_doc):
     ## 2. First Optimization - Create an index from each algorithm to file.
     ## 3. Use this mapping to go from algorithm to list of tutorial files in O(1) time.
 
+def source_links(algo_doc):
+    """
+    Get the source cpp and header links for the algorithm
+    """
+    # Source Directory and URL Prefix
+    SOURCE_PATH = '../../src/algorithms'
+    URL_PREFIX = 'https://github.com/MTG/essentia/tree/master/'
```
dbogdanov, line 156: Let's use prefix based on the source code used to build the documentation. Use the output of `git describe --long` instead of `master``.

dbogdanov (changes requested):

PR #1413, opened 2024-05-13
Add new algo audio2pitch

xaviliz:
This pull request add a new essntia algorithm, Audio2Pitch. It has been designed for real time pitch extraction.

- src/algorithms/tonal/audio2pitch.cpp
- src/algorithms/tonal/audio2pitch.h
- test/src/unittests/tonal/test_audio2pitch.py

Before we review everything it would be nice to have an idea of which test could be useful and any suggestions you could have. Kindly.


src/algorithms/tonal/audio2pitch.cpp
```diff
@@ -0,0 +1,130 @@
+#include "audio2pitch.h"
+
+using namespace essentia;
+using namespace standard;
+
+const char* Audio2Pitch::name = "Audio2Pitch";
+const char* Audio2Pitch::category = "Pitch";
+const char* Audio2Pitch::description = DOC("Extractor algorithm to compute pitch with several possible pitch algorithms, specifically targeted for real-time pitch detection on saxophone signals.");
```
dbogdanov, line 8: Currently, in all algorithms this DOC string starts with a sentence "This algorithm ..." summarizing what the algorithm does, which is then auto-parsed by doc generation script to create short descriptions for the [algorithm index](https://essentia.upf.edu/algorithms_reference.html). Let's keep it similar.

src/algorithms/tonal/audio2pitch.cpp
```diff
@@ -0,0 +1,130 @@
...
+  _loudnessAlgorithmName = parameter("loudnessAlgorithm").toString();
+  _tolerance = parameter("tolerance").toReal();
+  _pitchConfidenceThreshold = parameter("pitchConfidenceThreshold").toReal();
+  _loudnessThreshold = parameter("loudnessThreshold").toReal();
+
+  if (_maxFrequency > _sampleRate*0.5 ) {
+    throw EssentiaException("Audio2Pitch: Max frequency cannot be higher than Nyquist frequency");
+  }
+  if (_maxFrequency <= _minFrequency) {
+    throw EssentiaException("Audio2Pitch: Max frequency cannot be lower than min frequency");
```
dbogdanov, line 30: "lower or equal"

src/algorithms/tonal/audio2pitch.cpp
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  if (_pitchAlgorithmName != "pyin_fft" && _pitchAlgorithmName != "pyin") {
+    E_INFO("Audio2Pitch: 'pitchAlgorithm' = "<<_pitchAlgorithmName<<"\n");
```
dbogdanov, line 34: Why not to add the parameter value directly to the exception message?

src/algorithms/tonal/audio2pitch.cpp
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  if (_loudnessAlgorithmName == "loudness") {
+    _loudnessAlgorithm = AlgorithmFactory::create("Loudness");
+  }
+  else if (_loudnessAlgorithmName == "rms") {
+    _loudnessAlgorithm = AlgorithmFactory::create("RMS");
+  }
+  else {
+    E_INFO("Audio2Pitch: 'loudnessAlgorithm' = "<<_loudnessAlgorithmName<<"\n");
```
dbogdanov, line 60: Same here, we can provide the parameter value directly in the exception message.

src/algorithms/tonal/audio2pitch.cpp
```diff
@@ -0,0 +1,130 @@
...
+    throw EssentiaException("Audio2Pitch: Bad 'loudnessAlgorithm' parameter");
+  }
+
+  // switch between pyin and pyin_fft to propagate the weighting parameter
```
dbogdanov, line 64: Here we are directly using parameter strings for comparison, but `isSpectral` is used in other parts of the code. Is this because there could be multiple FFT-based algorithms in the future? If not, we could simplify and get rid of `isSpectral`.

dbogdanov (changes requested):

src/algorithms/tonal/audio2pitch.h
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    void declareParameters() {
+      declareParameter("sampleRate", "sample rate of incoming audio frames", "[8000,inf)", 44100);
+      declareParameter("frameSize", "size of input frame in samples", "[1,inf)", 1024);
+      declareParameter("minFrequency", "minimum frequency to detect in Hz", "[10,20000]", 60.f);
+      declareParameter("maxFrequency", "maximum frequency to detect in Hz", "[10,20000]", 2300.f);
+      declareParameter("pitchAlgorithm", "pitch algorithm to use", "{pyin,pyin_fft}", "pyin_fft");
+      declareParameter("loudnessAlgorithm", "loudness algorithm to use", "{loudness,rms}", "rms");
+      declareParameter("weighting", "string to assign a weighting function", "{default,A,B,C,D,Z}", "default");
```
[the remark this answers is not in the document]
xaviliz: By default PitchYinFFT apply a weighting function defined [here](https://github.com/MTG/essentia/blob/8f324eac55440bf02e370b903493a98c6d394660/src/algorithms/tonal/pitchyinfft.cpp#L37) we can think in another name though. Otherwise we can add a comment like `Default PitchYinFFt weighting function` ?

src/algorithms/tonal/audio2pitch.h
```diff
@@ -0,0 +1,80 @@
...
+      declareParameter("tolerance", "sets tolerance for peak detection on pitch algorithm", "[0,1]", 1.0f);
+      declareParameter("pitchConfidenceThreshold", "level of pitch confidence above which note ON/OFF start to be considered", "[0,1]", 0.25);
+      declareParameter("loudnessThreshold", "loudness level above which note ON/OFF start to be considered, in linear values", "[0,1]", 0.0031); // ~ -50dB
```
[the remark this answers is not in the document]
xaviliz: yes, decibels are easy to handle. I think we can make the conversion, it won't be an issue. Indeed we need dBs to for velocity conversion. Let me check which would be the impact.

src/algorithms/tonal/audio2pitch.cpp
xaviliz: well, this was because we started including other algorithms like PitchMelodia and in the python prototype we have more algorithms based on spectral processing. I think it might be removed if we don't plan to include others.

src/algorithms/tonal/audio2pitch.cpp
xaviliz: mmm... because operands are invalid in the exception message:
`../src/algorithms/tonal/audio2pitch.cpp:60:69: error: invalid operands to binary expression ('const char [39]' and 'std::string' (aka 'basic_string<char, char_traits<char>, allocator<char> >'))
    throw EssentiaException("Audio2Pitch: Bad 'loudnessAlgorithm' ="<<_loudnessAlgorithmName<<"\n");`
  Can you confirm that?

xaviliz: ok, I have seen how you use to do that in some other algorithm like here:
`throw EssentiaException(fullName(), ": Could not push 1 value, output buffer is full");`

src/algorithms/tonal/audio2pitch.h
```diff
@@ -0,0 +1,80 @@
...
+    void declareParameters() {
+      declareParameter("sampleRate", "sample rate of incoming audio frames", "[8000,inf)", 44100);
+      declareParameter("frameSize", "size of input frame in samples", "[1,inf)", 1024);
+      declareParameter("minFrequency", "minimum frequency to detect in Hz", "[10,20000]", 60.f);
+      declareParameter("maxFrequency", "maximum frequency to detect in Hz", "[10,20000]", 2300.f);
+      declareParameter("pitchAlgorithm", "pitch algorithm to use", "{pyin,pyin_fft}", "pyin_fft");
+      declareParameter("loudnessAlgorithm", "loudness algorithm to use", "{loudness,rms}", "rms");
+      declareParameter("weighting", "string to assign a weighting function", "{default,A,B,C,D,Z}", "default");
+      declareParameter("tolerance", "sets tolerance for peak detection on pitch algorithm", "[0,1]", 1.0f);
+      declareParameter("pitchConfidenceThreshold", "level of pitch confidence above which note ON/OFF start to be considered", "[0,1]", 0.25);
```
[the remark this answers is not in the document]
dbogdanov: To me "voice frame detection" is not as clear as "note ON/OFF". Also, using "above/below" is more consistent with "ON/OFF".

src/algorithms/tonal/audio2pitch.h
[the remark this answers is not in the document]
xaviliz: @dbogdanov I think this is still pending

src/algorithms/tonal/audio2pitch.cpp
dbogdanov: Let's get rid of `isSpectral` to simplify.

src/algorithms/tonal/audio2pitch.h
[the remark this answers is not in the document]
dbogdanov: Maybe we can call it `custom` and use a comment like you suggest: `Custom PitchYinFFt weighting function`

src/algorithms/tonal/audio2pitch.h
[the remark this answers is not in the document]
xaviliz: ok, I will go fix it now

src/algorithms/tonal/audio2pitch.h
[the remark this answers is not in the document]
dbogdanov: Should we use dB's or keep linear values?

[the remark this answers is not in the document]
xaviliz: This part is still pending. I think decibles is more consistent.

src/algorithms/tonal/audio2pitch.cpp
xaviliz: mmmm... interesting there is not E_INFO message in my last commits. Maybe do you need to pull my last changes?

src/algorithms/tonal/audio2pitch.h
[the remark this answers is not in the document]
xaviliz: Ok, it is done. But this implies to modify pitchyinfft algorithm.

xaviliz:
Okay! It is done. Please check from L192-197, I think this snippet would be nice to add it in all algorithms using AudioLoaders to pass properly the tests in osx.

test/src/unittests/tonal/test_audio2pitch.py
```diff
@@ -0,0 +1,227 @@
...
+        sample_rate = 44100
+        hopSize = 512
+        loudness_threshold = -80
+        filename = join(testdata.audio_dir, "recorded", "vignesh.wav")
+        if sys.platform == "darwin":
+            import soundfile as sf
+
+            audio, _ = sf.read(filename, dtype="float32")
+        else:
+            audio = MonoLoader(filename=filename, sampleRate=44100)()
```
dbogdanov, line 197: What is the rationale for using soundfile (libsndfile) instead of MonoLoader specifically on MacOS?

src/algorithms/tonal/audio2pitch.cpp
```diff
@@ -0,0 +1,128 @@
+#include "audio2pitch.h"
+#include "essentiamath.h"
+
+using namespace essentia;
+using namespace standard;
+
+const char* Audio2Pitch::name = "Audio2Pitch";
+const char* Audio2Pitch::category = "Pitch";
+const char* Audio2Pitch::description = DOC("This algorithm computes pitch with several possible pitch algorithms, specifically targeted for real-time pitch detection on audio signals.");
```
dbogdanov, line 9: Can simplify: "several possible" -> "various"

src/algorithms/tonal/audio2pitch.h
```diff
@@ -0,0 +1,77 @@
...
+    Real _loudnessThreshold;
+    Real _loudnessThresholdGain;
+    
+    bool isAboveThresholds(Real pitchConfidence, Real loudness);
+
+  public:
+    Audio2Pitch() {
+      declareInput(_frame, "frame", "the input frame to analyse");
+      declareOutput(_pitch, "pitch", "detected pitch in Hz");
+      declareOutput(_pitchConfidence, "pitchConfidence", "confidence of detected pitch from 0.0 - 1.0");
```
dbogdanov, line 41: we can put the range into parenthesis: "(from ... to ..)"

src/algorithms/tonal/pitchyinfft.cpp
```diff
@@ -37,7 +37,7 @@ static Real _weightMask[] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
 static const Real _weights[] = {-75.8, -70.1, -60.8, -52.1, -44.2, -37.5,
 	-31.3, -25.6, -20.9, -16.5, -12.6, -9.6, -7.0, -4.7, -3.0, -1.8, -0.8,
 	-0.2, -0.0, 0.5, 1.6, 3.2, 5.4, 7.8, 8.1, 5.3, -2.4, -11.1, -12.8,
-	-12.2, -7.4, -17.8, -17.8, -17.8}; // by default the original one is selected
+	-12.2, -7.4, -17.8, -17.8, -17.8}; // by default the original and custom one is selected
```
dbogdanov, line 40: This comment is a little fuzzy, maybe we can say something like "By default use custom weights designed specifically for the PitchYinFFT algorithm".

src/algorithms/tonal/audio2pitch.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+  if (frame.size() == 1) {
+    throw EssentiaException("Audio2Pitch: cannot compute the pitch of a frame of size 1");
+  }
+
+  if (_loudnessAlgorithmName == "loudness") {
+    _loudnessAlgorithm->input("signal").set(frame);
+    _loudnessAlgorithm->output("loudness").set(loudness);
+  }
+  else {
```
dbogdanov, line 99: maybe use `else if (_loudnessAlgorithmName == "rms")` to be more explicit (and also it is easier to read the logic)

src/algorithms/tonal/audio2pitch.h
```diff
@@ -0,0 +1,77 @@
...
+      declareParameter("sampleRate", "sample rate of incoming audio frames", "[8000,inf)", 44100);
+      declareParameter("frameSize", "size of input frame in samples", "[1,inf)", 1024);
+      declareParameter("minFrequency", "minimum frequency to detect in Hz", "[10,20000]", 60.0);
+      declareParameter("maxFrequency", "maximum frequency to detect in Hz", "[10,20000]", 2300.0);
+      declareParameter("pitchAlgorithm", "pitch algorithm to use", "{pyin,pyin_fft}", "pyin_fft");
+      declareParameter("loudnessAlgorithm", "loudness algorithm to use", "{loudness,rms}", "rms");
+      declareParameter("weighting", "string to assign a weighting function", "{custom,A,B,C,D,Z}", "custom");
+      declareParameter("tolerance", "sets tolerance for peak detection on pitch algorithm", "[0,1]", 1.0);
+      declareParameter("pitchConfidenceThreshold", "level of pitch confidence above/below which note ON/OFF start to be considered", "[0,1]", 0.25);
+      declareParameter("loudnessThreshold", "loudness level above/below which note ON/OFF start to be considered, in decibels", "[-inf,0]", -51.0);
```
dbogdanov, line 63: This assumes that there is the same good default threshold for both Loudness and RMS algorithms. Just wanted to double check if this is fine.

dbogdanov (changes requested):

test/src/unittests/tonal/test_audio2pitch.py
xaviliz: because at least in my machine MonoLoader and all algos using FFmpeg library are not working. So, it is the only way I had to run and pass all the tests.

xaviliz: Is the FFmpeg issue solved in others OS versions?

src/algorithms/tonal/audio2pitch.h
xaviliz: Right! I will do it.

src/algorithms/tonal/audio2pitch.h
xaviliz: ok, that's a good point. I have been using RMS all time. So, I will check the results with loudness with a unittest now.

xaviliz: No, it is not fine. Sorry, it is something we added in the initial definition but we never checked becasue RMS was working properly. Actually Loudness algorithm is not notmalized and cannot be mapped to decibles. So, I think LOUDNESS algorithm  can be removed. 

The same for PITCHYIN, in my preliminar tests in tests it is very slow to be run in real-time (at least in python). It is not an optimized solution when a large block is used like in this case. That's why we discard it. Maybe in cpp is working fine but I didn't test it properly. So, I think we can remove too.

test/src/unittests/tonal/test_audio2pitch.py
dbogdanov: Ideally, we should resolve the FFmpeg issue, however it is out of scope of this PR. We can remove this custom workaround (soundfile) and merge this PR, plus create a new GitHub issue related to the problem you found.

src/algorithms/tonal/audio2pitch.h
dbogdanov: To move forward, for now I propose to remove Loudness due to the inconsistency in thresholding, but keep PitchYin for now. Add a GitHub issue asking to benchmark PitchYin option in this algorithm for real-time.

src/algorithms/tonal/audio2pitch.cpp
```diff
@@ -6,7 +6,7 @@ using namespace standard;
 
 const char* Audio2Pitch::name = "Audio2Pitch";
 const char* Audio2Pitch::category = "Pitch";
-const char* Audio2Pitch::description = DOC("This algorithm computes pitch with several possible pitch algorithms, specifically targeted for real-time pitch detection on audio signals.");
+const char* Audio2Pitch::description = DOC("This algorithm computes pitch with various possible pitch algorithms, specifically targeted for real-time pitch detection on audio signals.");
```
dbogdanov, line 9: "possible" word is redundant

src/algorithms/tonal/audio2pitch.cpp
```diff
@@ -112,7 +112,7 @@ void Audio2Pitch::compute() {
     _spectrum->compute();
     _pitchAlgorithm->input("spectrum").set(spectrum);
   }
-  else {
+  else if (_pitchAlgorithmName == "pyin") {
```
dbogdanov, line 115: "pyin" is not a good name, because this abbreviation stands for pYIN, probabilistic YIN

I propose using `pitchyin`and `pitchyinfft` which directly correspond to our algorithm names PitchYin and PitchYinFFT. 

In the algorithm's DOC description, Mention these algorithms' names explicitly (e.g., "The algorithm internally uses pitch estimation with PitchYin (`pitchyin`) and PitchYinFFT (`pitchyinfft`)".

This way the documentation script will automatically add PitchYin and PitchYinFFT as related algorithms.

dbogdanov (changes requested):

test/src/unittests/tonal/test_audio2pitch.py
xaviliz: okay! let's do that.

src/algorithms/tonal/audio2pitch.cpp
xaviliz: alright, I will change it.

PR #1417, opened 2024-05-24
Add essentiamath

xaviliz:
Add some hertz, midi note number and note conversions:

- hz2midi
- midi2hz
- hz2cents
- cents2hz
- midi2note
- note2midi
- note2hz
- hz2note
- db2velocity
- velocity2db

### DIFF
--- a/src/essentia/essentiamath.h
+++ b/src/essentia/essentiamath.h
@@ -800,6 +800,16 @@ inline int note2midi(std::string note) {
   return midiNote;
 }
 
+inline std::string hz2note(Real hz, Real tuningFrequency) {
+    int midiNoteNumber = hz2midi(hz, tuningFrequency);
+    return midi2note(midiNoteNumber);
+}
+
+inline int note2hz(std::string note, Real tuningFrequency) {
+    int midiNoteNumber = note2midi(note);
+    return midi2hz(midiNoteNumber, tuningFrequency);
+}
+
 inline int db2velocity (Real decibels, Real hearingThreshold) {
   int velocity = 0;
   if (decibels > hearingThreshold) {
--- a/src/essentia/essentiamath.h
+++ b/src/essentia/essentiamath.h
@@ -738,6 +738,10 @@ inline Real hz2hz(Real hz){
   return hz;
 }
 
+inline Real cents2hz(Real frequencyB, Real cents) {
+  return frequencyB * powf(2, cents/1200)
+}
+
 inline Real hz2cents(Real frequencyA, Real frequencyB) {
   return 1200 * log2(frequencyA / frequencyB);
 }
--- a/src/essentia/essentiamath.h
+++ b/src/essentia/essentiamath.h
@@ -738,12 +738,12 @@ inline Real hz2hz(Real hz){
   return hz;
 }
 
-inline Real cents2hz(Real frequencyB, Real cents) {
-  return frequencyB * powf(2.0, cents / 1200.0);
+inline Real cents2hz(Real cents, Real referenceFrequency) {
+  return referenceFrequency * powf(2.0, cents / 1200.0);
 }
 
-inline Real hz2cents(Real frequencyA, Real frequencyB) {
-  return 1200 * log2(frequencyA / frequencyB);
+inline Real hz2cents(Real hz, Real referenceFrequency) {
+  return 1200 * log2(hz / referenceFrequency);
 }
 
 inline int hz2midi(Real hz, Real tuningFrequency) {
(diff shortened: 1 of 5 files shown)


src/essentia/essentiamath.h
```diff
@@ -738,8 +738,88 @@ inline Real hz2hz(Real hz){
...
 }
 
-inline Real hz2cents(Real hz) {
-  return 12 * std::log(hz/440)/std::log(2.) + 69;
+inline Real cents2hz(Real frequencyB, Real cents) {
+  return frequencyB * powf(2.0, cents / 1200.0);
+}
+
+inline Real hz2cents(Real frequencyA, Real frequencyB) {
+  return 1200 * log2(frequencyA / frequencyB);
```
dbogdanov, line 746: `frequencyA` vs `frequencyB` can be unclear to users. Maybe we could use a different naming for `frequencyB` like `referenceFrequency`?

src/essentia/essentiamath.h
```diff
@@ -738,8 +738,88 @@ inline Real hz2hz(Real hz){
...
+  const std::vector<std::string> ALL_NOTES { "A", "A#", "B", "C", "C#", "D", "D#", "E", "F", "F#", "G", "G#" };
+  int octave = note2octave(note);
+  std::string root = note2root(note);
+  int nNotes = ALL_NOTES.size();
+  int CIdx = 3;
+
+  int noteIdx = floor((octave - (CIdx + 1)) * nNotes);
+  int idx = 0;
+  for (int i = 0; i < nNotes; i++) {
+      if (ALL_NOTES[i] == root) {
```
dbogdanov, line 791: Break out of the loop once the matching  note is found.

src/python/globalfuncs.cpp
```diff
@@ -1001,6 +1162,18 @@ static PyMethodDef Essentia__Methods[] = {
   { "hz2bark",       hzToBark,         METH_O, "Converts a frequency in Hz to a bark band" },
   { "mel2hz",        melToHz,          METH_O, "Converts a mel band to frequency in Hz" },
   { "hz2mel",        hzToMel,          METH_O, "Converts a frequency in Hz to a mel band" },
+  { "midi2hz",       midiToHz,         METH_VARARGS, "Converts a midi note number to frequency in Hz" },
+  { "hz2midi",       hzToMidi,         METH_VARARGS, "Converts a frequency in Hz to a midi note number" },
+  { "hz2cents",      hzToCents,        METH_VARARGS, "Returns the cents distance between two frequencies in Hz" },
```
dbogdanov, line 1167: if we use the `referenceFrequency` (see my comment about), we should updates this description too.

src/python/globalfuncs.cpp
```diff
@@ -389,6 +389,167 @@ hzToMel(PyObject* notUsed, PyObject* arg) {
...
+  return PyLong_FromLong( int(midi) );
+}
+
+static PyObject*
+hzToCents(PyObject* notUsed, PyObject* args) {
+  // parse args to get Source alg, name and source alg and source name
+  vector<PyObject*> argsV = unpack(args);
+
+  if (argsV.size() != 2 || !PyFloat_Check(argsV[0]) || !PyFloat_Check(argsV[1])) {
+    PyErr_SetString(PyExc_TypeError, (char*)"expecting arguments (Real frequencyA, Real frequencyB)");
```
dbogdanov, line 425: Also here

src/python/globalfuncs.cpp
```diff
@@ -389,6 +389,167 @@ hzToMel(PyObject* notUsed, PyObject* arg) {
...
+  return PyFloat_FromDouble( int(cents) );
+}
+
+static PyObject*
+centsToHz(PyObject* notUsed, PyObject* args) {
+  // parse args to get Source alg, name and source alg and source name
+  vector<PyObject*> argsV = unpack(args);
+
+  if (argsV.size() != 2 || !PyFloat_Check(argsV[0]) || !PyFloat_Check(argsV[1])) {
+    PyErr_SetString(PyExc_TypeError, (char*)"expecting arguments (Real frequencyB, Real cents)");
```
dbogdanov, line 439: and here

dbogdanov (changes requested):

src/essentia/essentiamath.h
xaviliz: sure, that makes sense. I doubted with `tuningFrequency` but `referenceFrequency` works better. Thanks.

src/essentia/essentiamath.h
xaviliz: L796 is doing it by index. However it can be replaced by a `break` to make it more explicit. Thanks again to point it out.

xaviliz:
Thanks for the corrections and comments 🚀

PR #1434, opened 2024-07-19
Implementation of Direct Scale Transform Feature

dhunstack:
[feat] Complete Implementation of Direct Scale Transform
- Added parameters in header file
- Added configure function
- Added first order difference and matrix multiplication code

[feat] Add tests for Direct Scale Transform
Add the following tests:
 - Invalid Parameters Test
 - Invalid Inputs Test
 - Zero Input Test
 - Regression Test to compare with original Python implementation


dbogdanov:
Let's replace the `.txt` files with `.npy` binaries for numerical data in tests.

src/algorithms/standard/directscaletransform.cpp
```diff
@@ -0,0 +1,84 @@
...
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the Affero GNU General Public License
+ * version 3 along with this program.  If not, see http://www.gnu.org/licenses/
+ */
+#include "directscaletransform.h"
+#include "essentiamath.h"
+
+#define FOR(i,l,r) for(int i=l; i<r; i++)
```
dbogdanov, line 22: Let's avoid using this define for clarify.

src/algorithms/standard/directscaletransform.cpp
```diff
@@ -0,0 +1,84 @@
...
+using namespace standard;
+
+const char* DirectScaleTransform::name = "DirectScaleTransform";
+const char* DirectScaleTransform::category = "Standard";
+const char* DirectScaleTransform::description = 
+DOC("Computes Direct Scale Transform on a given matrix.\n" 
+    "This code was derived from the original DST paper:\n"
+    "[1] Williams, W. J., and E. J. Zalubas. Helicopter transmission fault detection via "
+    "time-frequency, scale and spectral methods. Mechanical systems and signal processing"
+    "14.4 (2000): 545-559.");
```
dbogdanov, line 34: Is there some reference implementation we could additionally cite here?

src/algorithms/standard/directscaletransform.cpp
```diff
@@ -0,0 +1,84 @@
...
+    const std::vector<std::vector<Real>>& matrix = _matrix.get();
+    std::vector<std::vector<Real>>& result = _result.get();
+
+    int N = matrix.size();
+    if (N == 0) throw EssentiaException("Input matrix is empty");
+
+    std::complex<Real> zi = std::complex<Real>(0, 1);
+    Real step = M_PI/log(N+1);
+    int num_rows = _C/step+1;       // Number of rows. +1 because need to include the first row which is the 0th multiple of step
+    int P = matrix[0].size();
```
dbogdanov, line 52: There's no guarantee on the consistency of the size of inner vectors, because we are using vector of vectors. Therefore, this should be checked.

dbogdanov (changes requested):

PR #1437, opened 2024-08-30
Add new algo audio2midi

xaviliz:
It includes header, cpp and unittests.


src/algorithms/tonal/audio2midi.cpp
```diff
@@ -0,0 +1,143 @@
...
+  _pitchConfidenceThreshold = parameter("pitchConfidenceThreshold").toReal();
+  _loudnessThreshold = parameter("loudnessThreshold").toReal();
+  _transposition = parameter("transpositionAmount").toInt();
+  _minOccurrenceRate = parameter("minOccurrenceRate").toReal();
+  _midiBufferDuration = parameter("midiBufferDuration").toReal();
+  _minNoteChangePeriod = parameter("minNoteChangePeriod").toReal();
+  _minOnsetCheckPeriod = parameter("minOnsetCheckPeriod").toReal();
+  _minOffsetCheckPeriod = parameter("minOffsetCheckPeriod").toReal();
+  
+  // define frameSize depending on sampleRate
```
dbogdanov, line 30: There is no clear predictable "continuous" behavior with respect to the specified sample rate.  The `sampleRate` can be any integer within the allowed range  (e.g., 12000, 22000, ...). What happens if the specified sample rate isn't one of the considered cases? To make the behavior more predictable, we could use frequency ranges for each `frameSize` to use.

src/algorithms/tonal/audio2midi.cpp
```diff
@@ -0,0 +1,143 @@
...
+      break;
+    default:
+      _frameSize = _fixedFrameSize;
+  }
+
+  _applyTimeCompensation = parameter("applyTimeCompensation").toBool();
+
+  if (_frameSize > _sampleRate * 0.5)
+  {
+    throw EssentiaException("Sax2Pitch: Frame size cannot be higher than Nyquist frequency");
```
dbogdanov, line 52: Legacy text from another application (`Sax2Pitch`)

dbogdanov, line 52: Clarify the text: different units of measurements compared. Unclear why this condition is required.

src/algorithms/tonal/audio2midi.cpp
```diff
@@ -0,0 +1,143 @@
...
+    break;
+  case 2:
+    noteOffMessage = _formatter->note_off(previousMidiNoteNumber, offsetTimeCompensation);
+    noteOnMessage = _formatter->note_on(midiNoteNumber, pitch, pitchConfidence, onsetTimeCompensation);
+    break;
+  default:
+    noteOnMessage = "";
+    noteOffMessage = "";
+    break;
+  }*/
```
dbogdanov, line 140: Redundant? Remove

src/algorithms/tonal/audio2midi.cpp
```diff
@@ -0,0 +1,143 @@
+#include "audio2midi.h"
+
+using namespace std;
+using namespace essentia;
+using namespace standard;
+
+const char *Audio2Midi::name = "Audio2Midi";
+const char *Audio2Midi::category = "Pitch";
+const char *Audio2Midi::description = DOC("Wrapper around Audio2Pitch and Pitch2Midi for real time application");
```
dbogdanov, line 9: The documentation does not clarify that the algorithm has a state that is used to estimate note on/off events based on consequent compute() call. We could explicitly mention this. Also related to DOC in Pitch2Midi.

test/src/unittests/tonal/test_audio2midi.py
```diff
@@ -0,0 +1,312 @@
...
+        hop_size = 64
+        loudness_threshold = -40
+        pitch_confidence_threshold = 0.25
+        min_frequency = 103.83
+        max_frequency = 659.26
+        midi_buffer_duration = 0.05
+        min_note_change_period = 0.03
+        n_notes_tolerance = 0
+        onset_tolerance = 0.01
+        midi_note_tolerance = 0
```
dbogdanov, line 94: Too verbose. Why not specifying this directly inside `runARealCase()`?

test/src/unittests/tonal/test_audio2midi.py
```diff
@@ -0,0 +1,312 @@
...
+            onset_tolerance=onset_tolerance,
+            midi_note_tolerance=midi_note_tolerance,
+        )
+
+    def testARealCaseWithDMinorScale(self):
+        frame_size = 8192
+        sample_rate = 48000
+        hop_size = 64
+        loudness_threshold = -40
+        pitch_confidence_threshold = 0.25
```
dbogdanov, line 122: Same here. Specify directly in `runARealCase()`.

test/src/unittests/tonal/test_audio2midi.py
```diff
@@ -0,0 +1,312 @@
...
+        min_frequency = 103.83
+        max_frequency = 659.26
+        midi_buffer_duration = 0.05
+        min_note_change_period = 0.02
+        min_onset_period = 0.015
+        min_offset_period = 0.015
+        min_occurrence_rate = 0.05
+        n_notes_tolerance = 0
+        onset_tolerance = 0.01
+        midi_note_tolerance = 0
```
dbogdanov, line 167: Same here. Complicates reading.

test/src/unittests/tonal/test_audio2midi.py
```diff
@@ -0,0 +1,312 @@
...
+        min_frequency: float,
+        midi_buffer_duration: float,
+        min_note_change_period: float,
+        min_onset_period: float = 0.075,
+        min_offset_period: float = 0.2,
+        min_occurrence_rate: float = 0.5,
+        n_notes_tolerance: int = 0,
+        onset_tolerance: float = 0.01,
+        offset_tolerance: float = 0.05,
+        midi_note_tolerance: int = 0,
```
dbogdanov, line 212: Do we need these default values, or they are always explicitly specified in the calls to this function (which would be nicer)?

test/src/unittests/tonal/test_audio2midi.py
```diff
@@ -0,0 +1,312 @@
...
+    ):
+        filename = join(testdata.audio_dir, audio_path)
+        if sys.platform == "darwin":
+            import soundfile as sf
+
+            audio, _ = sf.read(filename, dtype="float32")
+            if audio.ndim > 1:
+                audio = audio[:, 0]
+        else:
+            audio = MonoLoader(filename=filename, sampleRate=sample_rate)()
```
dbogdanov, line 222: Use `MonoLoader` for both cases for consistency and also to avoid additional dependency.

test/src/unittests/tonal/test_audio2midi.py
```diff
@@ -0,0 +1,312 @@
...
+            minNoteChangePeriod=min_note_change_period,
+            minOnsetCheckPeriod=min_onset_period,
+            minOffsetCheckPeriod=min_offset_period,
+            pitchConfidenceThreshold=pitch_confidence_threshold,
+            loudnessThreshold=loudness_threshold,
+            maxFrequency=max_frequency,
+            minFrequency=min_frequency,
+            minOccurrenceRate=min_occurrence_rate,
+        )
+        print(a2m.parameterNames())
```
dbogdanov, line 240: Avoid verbose prints when unnecessary (may clutter the output when running all tests).

test/src/unittests/tonal/test_audio2midi.py
```diff
@@ -0,0 +1,312 @@
...
+        self.assertEqual(time_compensation.tolist(), array([], dtype=float32).tolist())
+
+    def assessNoteList(
+        self,
+        reference_path: str,
+        estimated: list,
+        n_notes_tolerance: int = 0,
+        onset_tolerance: float = 0.01,
+        offset_tolerance: float = 0.01,
+        midi_note_tolerance: int = 0,
```
dbogdanov, line 47: This can be hardcoded inside asserts directly

test/src/unittests/tonal/test_audio2midi.py
```diff
@@ -0,0 +1,312 @@
...
+
+        print("\ndiffs")
+        print(array(estimated) - expected_notes[:, 1:])
+
+        # estimate the number of notes for expected and detected
+        n_detected_notes = len(estimated)
+        n_expected_notes = len(expected_notes)
+
+        # estimate the onset error for each note and estimate the mean
+        onset_mse = mean(
```
dbogdanov, line 62: This check the mean error, however, are there any limits on individual error we should check?

dbogdanov (changes requested):

src/algorithms/tonal/audio2midi.cpp
xaviliz: I think is not needed if we define the frameSize internally, it is legacy code. Initially it was a sanity check when the frameSize was a pameter. I will remove this snippet.

test/src/unittests/tonal/test_audio2midi.py
xaviliz: Right, we can remove all prints. I left them just to understand the outputs.

test/src/unittests/tonal/test_audio2midi.py
xaviliz: No, it cannot bacause it is inside runARealCase() which receive different values for those parameters. If we hardcoded some tests will fail. I can remove the default value, that would be okay.

test/src/unittests/tonal/test_audio2midi.py
xaviliz: Off course, we can add a local assessment. I was checking other tests and they used a similar strategy.

test/src/unittests/tonal/test_audio2midi.py
xaviliz: Right, but I running tests in OSX. So, I need it to run them. Once we resolve all comments I might remove it. Let's fix the other parts and I will do it.

test/src/unittests/tonal/test_audio2midi.py
xaviliz: Please resolve or suggest any others changes.

test/src/unittests/tonal/test_audio2midi.py
dbogdanov: To be able to merge this now, let's get rid of the soundfile dependency and let the test temporarily fail on MacOS.

dbogdanov (changes requested):

PR #1426, opened 2024-07-01
[feat] Convert the documentation to Read The Docs template

dhunstack:
Following changes were made in this commit -
 - Sphinx's `conf.py` file was regenerated. Old configuration file is in `conf_old.py`
 - `custom.css` was added to set sidebar colors
 - `documentation.rst` page was added to serve as the home page for docs. Replaces the old `documentation.html`.
 - Some `rst` files were modified to reduce the number of sidebar pages

Screenshot 2024-07-01 at 16 02 40
Screenshot 2024-07-01 at 16 05 43

PR #1444, opened 2024-10-17
[fix] Add theme dependency in docs workflow

dhunstack:
Add the package `sphinx-rtd-theme` in `build-docs.yml` workflow file


dbogdanov:
Already addressed in https://github.com/MTG/essentia/pull/1426.

PR #1445, opened 2024-10-21
Add tempo cnn extractor

xaviliz:
Code for Tempo CNN extractor.


src/examples/standard_tempo_cnn.cpp
```diff
@@ -0,0 +1,100 @@
+/*
```
dbogdanov, line 1: Rename the file to `standard_tempocnn` so that it is easier to match with the name of the algorithm.

dbogdanov (changes requested):
All looks good. Just a tiny comment on the extractor name.

src/examples/standard_tempo_cnn.cpp
xaviliz: it is done!

PR #1460, opened 2025-01-17
Added unsupported python versions note in install instructions

Pineapple217:
Python 3.12 and higher won't install/work because of the removal of the `imp` package in `python 3.12`, [source](https://docs.python.org/3.12/library/imp.html). This just adds a small note to warn people of this.


philippludwig:
@nolan-veed The guide at https://essentia.upf.edu/installing.html states "2.1 beta5 is the current stable version recommended to install." - when you choose this tag, you run into exactly this problem. On master, this does not occur.

PR #1482, opened 2025-08-02
Fix 'cannot seek vector iterator after end' debug assertion failure

wo80:
This pull request fixes a small issue with the order of operations in the `Key::shiftPcp()` method.

The issue occurs on Windows with MSVC compiler and leads to a 'cannot seek vector iterator after end' debug assertion failure at runtime.

The problem was reported on my fork https://github.com/wo80/essentia/issues/7


wo80:
No interest, it seems.

PR #1500, opened 2026-02-19
fix: harden AudioLoader against SIGSEGV from corrupt/malformed audio frames

xiaden:
This PR Adds defensive guards in both decodePacket() and flushPacket() to prevent segmentation faults when processing corrupted or malformed audio files:

- Check for null frame data pointers before memcpy/swr_convert
- Validate channel count matches expected layout (prevents swr misuse)
- Sanity-check nb_samples to prevent integer overflow in buffer size calc
- Skip frames that would overflow FFMPEG_BUFFER_SIZE instead of clamping
- Remove redundant std::min() clamping that masked the overflow silently

All guards emit E_WARNING with diagnostic details (codec name, sample count, channel count) and skip the offending frame rather than crashing.

### Behavior

 - Valid audio: No observable change (A/B tested in bulk ML tagging workload)
 - Partially corrupted audio: Frames may now be skipped instead of processed
 - Fully corrupted frames that previously triggered SIGSEGV: Now handled safely

### Notes

 - This implementation opts to skip invalid frames rather than abort processing. If a different failure strategy is preferred, I’m happy to adapt.

 - The changes were implemented with AI assistance, and I’m not deeply familiar with all invariants in this codebase, so I’d appreciate careful review, particularly around buffer handling and swr assumptions.

### DIFF
--- a/src/algorithms/io/audioloader.cpp
+++ b/src/algorithms/io/audioloader.cpp
@@ -369,23 +369,59 @@ void AudioLoader::flushPacket() {
         }
 
         // got a frame -> convert to floats as in decodePacket()
+        // Guard: null frame data — corrupt decoder output
+        if (!_decodedFrame->data[0]) {
+            E_WARNING("AudioLoader: flushed frame has null data pointer, skipping frame");
+            continue;
+        }
+        // Guard: channel count mismatch — swr was configured for _nChannels in openAudioFile()
+        if (_decodedFrame->ch_layout.nb_channels != _nChannels) {
+            ostringstream msg;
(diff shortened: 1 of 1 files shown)


xiaden:
Hi maintainers, small follow-up on this PR.

I've now been using this patch actively in Nomarr for bulk audio processing, and can confirm it's behavior has been stable across the roughly 430k files I've processed since implementing.

Is this failure-handling approach acceptable for `AudioLoader`?  I'm happy to rebase, add validation, or rework this patch if needed.

PR #1501, opened 2026-02-19
feat: lazy TF session init - defer GPU probe until model load

xiaden:
Defer TensorFlow graph/session creation from the constructor to configure().

## Description

Previously, TensorFlow session creation occurred in the constructor, causing GPU device probing (and associated log output) when essentia.standard was imported — even if no TensorFlow-based algorithms were used.

This change defers TF object creation until TensorflowPredict::configure() is called with a valid model path.

### Behavior Change

 - Before: GPU initialization triggered at import essentia.standard
 - After: GPU initialization occurs only when a TensorFlow model is actually configured

The session is still created before compute() runs, so runtime behavior for valid usage remains unchanged.
The only observable difference is that GPU probing logs now appear at model load time instead of import time.

### Notes

Adds guards in destructor and reset() to handle uninitialized TF objects safely.
Prevents unnecessary GPU probing when TensorFlow algorithms are registered but never used.

The changes were implemented with AI assistance, and I’m not deeply familiar with all invariants in this codebase, so I’d appreciate careful review.

### DIFF
--- a/src/algorithms/machinelearning/tensorflowpredict.cpp
+++ b/src/algorithms/machinelearning/tensorflowpredict.cpp
@@ -80,6 +80,14 @@ void TensorflowPredict::configure() {
   // Do not do anything if we did not get a non-empty model name.
   if ((_savedModel.empty()) and (_graphFilename.empty())) return;
 
+  // Lazy initialization: Create TF objects on first real configure() call.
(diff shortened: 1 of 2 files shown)


palonso:
Hi @xiaden,
I've built and tested the PR locally, and it looks good to me.

@dbogdanov, we can merge it.

PR #1510, opened 2026-05-04
Updates in FreesoundExtractor 

ffont:
This PR includes some updates in the `FreesoundExtractor`:

* Removed check for completely silent audio in FreesoundExtractor. There are files in Freesound which will be complete silence or almost silence, so it is fine to provide results for them as well. 
* Add computation of the Freesound similarity space vectors, based on a normalization of the low level descriptors and PCA transformation. Coefficients for such operations are hard-coded in a separate (huge) header file. 
* Add some post-processing of computed descriptors that will output midi note numbers, names, and some other minor stuff (in a separate `fs.x` namespace). This is used in Freesound when indexing the descriptors. We used to do that post-processing in Freesound code, but doing it here it is useful as people that uses the `FreesoundExtractor` to analyze sounds locally, gets descriptors which are directly compatible with those indexed in Freesound.


ffont:
I think the implementation is probably fine, but I'm a bit concerned about the size of the hard-coded literals in `src/essentia/utils/extractor_freesound/FreesoundSimSpaceDescriptorsData.h`.

Also note that this includes Pablo's changes to make builds work. Maybe better to merge these first to master so the diff here makes more sense.

dbogdanov:
@ffont: changes by @palonso  were merged in a separate PR, but the diff is not updated...

ffont:
Ja he fet merge back de master i es veuen els canvis bé!

PR #1514, opened 2026-05-24
Add Windows (AMD64) Python wheel support (MSVC + vcpkg)

joseluu:
## Summary

Adds Windows AMD64 Python wheel support via cibuildwheel, building with MSVC (VS 2022) and vcpkg x64-windows-static-md.

Windows users currently cannot \`pip install essentia\` — they are directed to build from source. This PR adds both the CI infrastructure and all required C++ compatibility fixes to build and publish Windows wheels for Python 3.12+.

## What this PR does

**CI / build system:**
- \`cibuildwheel.toml\`: Windows AMD64 wheel build via GitHub Actions \`windows-latest\`, using vcpkg x64-windows-static-md (static libs + DLL CRT compatible with CPython)
- \`wscript\`: Add MSVC-specific flags: \`/std:c++14\`, \`/W2\`, \`/EHsc\`, \`/bigobj\`, \`/MD\` (for both C and C++ so all objects match the DLL CRT required by .pyd files)
- \`wscript\`: Add \`NOMINMAX\` and \`_USE_MATH_DEFINES\` Windows defines
- \`src/wscript\`: Filter out \`-lm\` on Windows (\`m.lib\` does not exist; math is in the Windows CRT)
- \`src/python/wscript\`: Replace removed \`distutils.sysconfig\` with \`sysconfig\` module (Python 3.12)

**C++ source fixes:**
- \`roguevector.h\`: Replace \`_Myfirst()\`/\`_Mylast()\`/\`_Myend()\` (MSVC internals removed in VS 2022) with raw pointer cast
- \`yamlast.h\`: Define \`YAML_DECLARE_STATIC\` before \`<yaml.h>\` — prevents libyaml headers from using \`__declspec(dllimport)\` which conflicts with static \`yaml.lib\` from vcpkg (LNK2019)
- \`tensornormalize.cpp\` (MINMAX path): Materialize \`Eigen::Tensor\` \`broadcast()\` expressions into concrete \`Tensor<Real>\` before combining — MSVC \`/O2\` hangs for hundreds of CPU-seconds on deeply nested Eigen expression templates
- \`humdetector.cpp\`, \`vectorrealtotensor.cpp\`, \`key.cpp\`: Replace C++ alternative tokens (\`not\`, \`or\`) with \`!\`, \`||\` — MSVC requires \`<iso646.h>\` for these without \`/Za\`

**Python binding fixes:**
- \`typedefs.h\`: Rename enum \`Edt\` member \`BOOL → EDT_BOOL\` to avoid conflict with \`typedef int BOOL\` from Windows SDK \`minwindef.h\` (C2365)
- \`parsing.cpp\`, \`pyalgorithm.cpp\`, \`pystreamingalgorithm.cpp\`: Update all \`BOOL\` → \`EDT_BOOL\` in switch/call sites
- \`tensorreal.cpp\`, \`vectortensorreal.cpp\`: Replace VLA \`npy_intp dims[nd]\` with \`dims[TENSORRANK]\` — MSVC does not support C99 variable-length arrays; \`TENSORRANK\` is the compile-time constant 4

## Test plan

- [x] Tested locally on Windows 11 x64 with VS 2022 Build Tools, Python 3.12, vcpkg x64-windows-static-md
- [x] All imports succeed on a clean Windows environment (no MSVC env vars needed at runtime):
  ```python
  import essentia
  import essentia.standard
  import essentia.streaming
  from essentia.standard import MonoLoader
  ```
- [ ] CI wheels build on GitHub Actions (pending review)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build-wheels-cibuildwheel.yml
+++ b/.github/workflows/build-wheels-cibuildwheel.yml
@@ -7,6 +7,7 @@ jobs:
     name: ${{ matrix.os }} / ${{ matrix.config }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-24.04, macos-15-intel, macos-15, windows-latest]
         config: [cibuildwheel, cibuildwheel-tensorflow]
--- a/.github/workflows/build-wheels-cibuildwheel.yml
+++ b/.github/workflows/build-wheels-cibuildwheel.yml
@@ -8,8 +8,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-24.04, macos-15-intel, macos-15]
+        os: [ubuntu-24.04, macos-15-intel, macos-15, windows-latest]
         config: [cibuildwheel, cibuildwheel-tensorflow]
+        exclude:
+          # TensorFlow Windows support is not yet implemented.
+          - os: windows-latest
+            config: cibuildwheel-tensorflow
 
     steps:
       - uses: actions/checkout@v6
@@ -19,6 +23,12 @@ jobs:
       - name: Fetch release tags from GitHub
         # Workaround for https://github.com/actions/checkout/issues/290
         run: git fetch --tags --force
+
+      - name: Set up MSVC environment (Windows only)
+        if: runner.os == 'Windows'
+        uses: ilammy/msvc-dev-cmd@v1
+        with:
+          arch: x64
 
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.2.1
--- a/cibuildwheel.toml
+++ b/cibuildwheel.toml
@@ -95,15 +95,16 @@ skip = ["*t-*"]
 # PKG_CONFIG_PATH (semicolon-separated on Windows):
 #   - C:/essentia-install  : libessentia installed by before-all
 #   - vcpkg installed dir  : fftw3, ffmpeg, taglib, chromaprint, etc.
-environment = { PROJECT_NAME="essentia", ESSENTIA_PROJECT_NAME="${PROJECT_NAME}", ESSENTIA_WHEEL_SKIP_3RDPARTY=1, ESSENTIA_WHEEL_ONLY_PYTHON=1, PKG_CONFIG_PATH="C:/essentia-install/lib/pkgconfig;$VCPKG_INSTALLATION_ROOT/installed/x64-windows-static-md/lib/pkgconfig" }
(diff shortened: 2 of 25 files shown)


wo80:
@joseluu Did you test the RogueVector fix for Visual Studio? I tested this locally with the code from my [fork](https://github.com/wo80/essentia/tree/cmake) (running the C++ googletest project). I'm getting a bunch of errors, like:

```
D:\projects\native\MTG\essentia_20260820>ctest --test-dir build --output-on-failure -C Debug
Test project D:/projects/native/MTG/essentia_20260820/build
        Start   1: AudioLoader.SimpleLoad
  1/132 Test   #1: AudioLoader.SimpleLoad ...............................Exit code 0xc0000409***Exception:   0.16 sec
Note: Google Test filter = AudioLoader.SimpleLoad
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from AudioLoader
[ RUN      ] AudioLoader.SimpleLoad
C:\Program Files\Microsoft Visual Studio\18\Community\VC\Tools\MSVC\14.51.36231\include\vector(1939) : 
    Assertion failed: vector subscript out of range
```

joseluu:
@wo80 Good catch, thanks for the detailed repro — you were right, and I found the root cause.

The VS2019+ branch assumed `std::vector<T>`'s three internal pointers (`_Myfirst`, `_Mylast`, `_Myend`) always start at offset 0 of the object — that's only true when iterator debugging is disabled (`_ITERATOR_DEBUG_LEVEL == 0`, the Release default). In a Debug build (`_ITERATOR_DEBUG_LEVEL == 2` by default), MSVC's STL prepends a hidden `_Container_proxy*` before those three pointers, so my offset-0 write was clobbering the proxy pointer instead of `_Myfirst`. That corrupts the vector object, and the STL's debug checks (correctly) trip the "vector subscript out of range" assertion the next time the vector is touched — exactly what you hit on `AudioLoader.SimpleLoad`.

Confirmed the layout difference directly: `sizeof(std::vector<int>)` is 24 bytes in Release and 32 bytes in Debug on VS2022 (14.43.34808) — the extra 8 bytes is that proxy pointer.

Fix pushed in `6fdfb76f`: skip that extra pointer slot when `_ITERATOR_DEBUG_LEVEL != 0`. I validated it against the real header (not just a standalone repro) in both `/MD` and `/MDd` builds — construction, iteration, and `at()` bounds-checked access all work now in Debug, and I reproduced your exact failure mode (corrupted `size()`) by reverting just the offset compensation, to confirm the fix actually addresses the mechanism you found rather than just moving the symptom.

I also split the Windows branch on `_MSC_VER` (`<=1916` keeps the old `_Myfirst()/_Mylast()/_Myend()` accessor-based path, `>1916` gets the new offset-aware raw-pointer path) instead of the bare `OS_WIN32` check that was there before, so this doesn't silently miscompile on a non-MSVC `OS_WIN32` target.

I don't have your `cmake`/googletest setup running here, so if you get a chance to re-test against `6fdfb76f` (including the `_Container_proxy` path — i.e. an actual Debug `ctest` run, not just Release) I'd appreciate confirmation. Thanks again for flagging this instead of it landing as a silent Debug-only footgun for whoever builds mainline with `/MDd`.

wo80:
Googletest reporting
```
100% tests passed out of 132
```
for both release and debug builds.

PR #1523, opened 2026-08-13
build: use sysconfig instead of the removed distutils in the python bindings

SimplicityGuy:
`distutils` was deprecated by [PEP 632](https://peps.python.org/pep-0632/) and **removed from the standard library in Python 3.12**, so `python waf configure --with-python` fails outright on any modern interpreter:

```
File "src/python/wscript", line 6, in <module>
    import distutils.sysconfig
ModuleNotFoundError: No module named 'distutils'
```

Both uses are `distutils.sysconfig.get_python_lib()`, whose documented replacement is `sysconfig.get_paths()['purelib']`. `sysconfig` has been in the standard library since Python 3.2, so this raises no minimum version for anyone still on older interpreters.

Tested by building the bindings from source on Python 3.14.

Two lines changed. Signed off per the DCO in the contribution guide; under the 20-line CLA threshold.

### DIFF
--- a/src/python/wscript
+++ b/src/python/wscript
@@ -3,8 +3,8 @@
 
 from __future__ import print_function
 
-import distutils.sysconfig
 import os
+import sysconfig
 
 
 def options(ctx):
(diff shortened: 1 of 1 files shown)


SimplicityGuy:
Note on CI: the `cibuildwheel` jobs on this PR fail in `setup.py`'s `get_version()` with `IndexError: list index out of range`, **before any compilation** and unrelated to this change.

The cause is that `git describe --always --tags` emits a bare sha when no tag is reachable — the normal state of a fork, since GitHub does not copy tags — and `split('-')[-2]` then indexes past the end. It affects every PR opened from a fork.

Fixed separately in #1525, which is independent of this change and can be merged on its own. This PR's CI should go green once that lands (or on a rebase after it).

Issue #771, opened 2018-07-31
Add startTime/endTime seeking to AudioLoader, MonoLoader and EasyLoader


dbogdanov:
EasyLoader has ```startTime``` and ```endTime``` parameters, however those are implemented inefficiently: first the complete audio stream is loaded and then the required time segment is trimmed. This becomes a problem on very long tracks (e.g., a 1h recording from which a short segment should be extracted). 

This problem can be avoided by implementing the seeking functionality in AudioLoader (and related MonoLoader and EasyLoader).

https://0xdeafc0de.wordpress.com/2014/07/21/seeking-with-ffmpeg/
https://ffmpeg.org/pipermail/libav-user/2016-September/009608.html
https://stackoverflow.com/questions/31854406/precise-seeking-with-ffmpeg

The seeking for startTime can be implemented inside [openAudioFile()](https://github.com/MTG/essentia/blob/master/src/algorithms/io/audioloader.cpp#L53). We can then track the endTime inside process().

dbogdanov:
Also, this will be useful for when one wants to extract a stereo segment (```EasyLoader```'s output is mono.)

SimplicityGuy:
Is anyone currently working on this? If not, I'd like to take a run at it.

Context: I'm using `MonoLoader`/`EasyLoader` to analyze long audio (multi-hour recordings) in bounded windows, and the lack of seek means each window decode reads the stream from byte 0. That makes the per-file cost `O(n_windows x duration)` instead of `O(duration)`. A `Trimmer` at the head of the network does shut the decode down early, so an `endTime`-style bound is partly achievable today, but `startTime` has no equivalent — there's no way to skip the leading portion.

The plan I have in mind follows the sketch in the original post:

1. `av_seek_frame` / `avformat_seek_file` on `_demuxCtx` inside `openAudioFile()` for `startTime`, plus a `avcodec_flush_buffers()` on `_audioCtx`.
2. Because seeking in compressed audio lands on a packet boundary and not on a sample, seek to a target *before* the requested position and then decode-and-discard forward to the exact sample. Without this, the output isn't sample-accurate and window-based analysis stops being reproducible. I'd want the seeked output to be bit-identical to the current decode-from-zero-and-trim path, and would write the test that way.
3. Track `endTime` in `process()` and signal end-of-stream at the boundary.
[...]

PR #1524, opened 2026-08-13
fix(audioloader): set pkt_timebase so decoded frame timestamps are correct

SimplicityGuy:
`AVCodecContext::pkt_timebase` tells libavcodec the time base of the packets it is being fed. `AudioLoader` never sets it, so it stays at its `0/1` default.

libavcodec uses it whenever it has to move a timestamp by a sample count — most importantly when it discards a codec's initial padding (Opus pre-skip, AAC/MP3 encoder delay) and then advances `frame->pts` by the number of samples skipped. With an invalid `pkt_timebase` that adjustment is meaningless, so `frame->pts` does not correspond to the samples actually returned.

Nothing in `AudioLoader` consumed `frame->pts` before, which is why this has been invisible. It becomes load-bearing as soon as timestamps are used to locate the decoder in the stream — e.g. any seeking implementation (#771), which is how I found it.

**Measured:** on an Ogg Opus file the error is exactly the 312-sample (6.5 ms) Opus pre-skip. With this fix, decoded frame timestamps line up with the samples returned, and a seek-and-trim path becomes bit-identical to decode-and-discard on Ogg/Opus; without it, it is off by exactly the pre-skip.

Six lines added, none removed. Signed off per the DCO in the contribution guide; under the 20-line CLA threshold.

### DIFF
--- a/src/algorithms/io/audioloader.cpp
+++ b/src/algorithms/io/audioloader.cpp
@@ -115,6 +115,13 @@ void AudioLoader::openAudioFile(const string& filename) {
         throw EssentiaException("AudioLoader: Could not copy codec parameters");
     }
 
+    // Tell the decoder what time base the packets we feed it are in. libavcodec needs this to
+    // report frame timestamps correctly -- in particular it advances frame->pts by however many
+    // samples it skipped (codec delay / pre-skip), and that adjustment is expressed in
+    // pkt_timebase. Left unset the adjustment is computed against an invalid time base, so
+    // frame->pts is wrong for every format that carries an encoder delay.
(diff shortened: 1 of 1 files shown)


SimplicityGuy:
Note on CI: the `cibuildwheel` jobs on this PR fail in `setup.py`'s `get_version()` with `IndexError: list index out of range`, **before any compilation** and unrelated to this change.

The cause is that `git describe --always --tags` emits a bare sha when no tag is reachable — the normal state of a fork, since GitHub does not copy tags — and `split('-')[-2]` then indexes past the end. It affects every PR opened from a fork.

Fixed separately in #1525, which is independent of this change and can be merged on its own. This PR's CI should go green once that lands (or on a rebase after it).

PR #1527, opened 2026-08-27
Add startTime/endTime seeking to AudioLoader, MonoLoader and EasyLoader

### LINKED ISSUE #771: Add startTime/endTime seeking to AudioLoader, MonoLoader and EasyLoader

SimplicityGuy:
Closes #771 — startTime/endTime seeking in AudioLoader and MonoLoader, and EasyLoader/EqloudLoader switched from decode-everything-and-trim to seeking.

## Problem

EasyLoader and EqloudLoader have exposed `startTime`/`endTime` for years, but implement them by decoding the **entire** stream from byte 0 and trimming downstream. Extracting a slice costs O(total duration) no matter how short the slice is, so analysing a long recording in N windows costs O(N × duration) instead of O(duration).

## What changed

**AudioLoader** (owns the demuxer and codec contexts, so the seek belongs here):
- `openAudioFile()` finishes by calling `seekToStartTime()`, which issues `av_seek_frame(..., AVSEEK_FLAG_BACKWARD)` at (startTime − 0.5 s) and flushes the decoder and resampler.
- `process()` stops and reports FINISHED once endTime has been produced, instead of reading on to EOF.
- `copyFFmpegOutput()` trims the partial frames at both ends of the slice.

**MonoLoader** forwards both parameters.

**EasyLoader / EqloudLoader** hand `startTime`/`endTime` to the loader instead of the Trimmer. Same names, same defaults, same ranges, same meaning — and the loader converts seconds to samples with Trimmer's own truncation rule, so it selects exactly the samples the Trimmer used to select.

## Sample accuracy

A compressed stream can only be seeked to a packet boundary, and the packet it lands on is not independently decodable (MP3 bit reservoir; every MDCT codec needs the previous window). So the seek deliberately lands **half a second early**, decodes forward discarding output until the exact requested sample, and only then starts producing. The exact sample is located from the first decoded frame's pts, not assumed from the packet boundary.

Verified against the decode-from-zero-and-trim path on synthetic audio at several offsets:

| Container | Result |
|---|---|
| wav / mp3 / flac | bit-identical |
| ogg (Opus) | bit-identical, apart from one 2⁻²⁴ ULP at one offset |
| m4a (AAC) | bit-identical when the encoder did not use PNS |

The AAC caveat is a property of the codec, not the patch: Perceptual Noise Substitution synthesises noise bands from a PRNG whose state depends on the whole decode history, so no amount of preroll can reproduce it — confirmed by the residual being completely insensitive to preroll length (0.05 s to 10 s) and vanishing entirely when the file is re-encoded with `-aac_pns 0`.

## One case deliberately stays on the old path

A read that also **resamples**. libsamplerate's output depends on how much input it has already consumed (filter history plus a phase accumulator), so a converter started at startTime does not reproduce one started at zero — measured at about −49 dB relative RMS, and not removable by any bounded preroll. Making a resampled slice seek would silently change what every existing caller gets, so when `sampleRate` differs from the file's own rate, both composites keep decoding and trimming as before. Callers who want the win there and can accept the residual have MonoLoader's own `startTime`/`endTime`.

## Failure is soft everywhere it can be

A seek that fails, or a stream with no usable timestamps, falls back to decoding from the beginning and discarding — exactly today's behaviour, slow but correct — so no format that works now can break.

## Performance

On a 3600 s MP3, reading one 30 s window 1800 s in with EasyLoader at the file's own rate: **2.07 s before, 0.085 s after**. Reading the second half of the file now costs 53% of reading all of it instead of 92%.

## Tests

Extends the four existing loader test files, using only fixtures already in `test/audio`. Every accuracy test compares against the same reference: read the whole file, cut the slice out afterwards. Boundary cases covered: startTime 0, endTime past/at EOF, startTime past EOF (empty, not an error), an empty slice, an inverted slice (an error), and a file shorter than the seek preroll. Mutation-checked: removing the preroll fails 8 tests, rounding instead of truncating fails 4, ignoring endTime fails 21.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MY2RbcC317YTPuJsiGJrvd

## Opus pre-skip regression test (added)

The branch now also carries `testSeekOggOpus`: seeking relies on the first decoded frame's `pts`, which is correct only because #1524 set `AVCodecContext::pkt_timebase` (see #1531) — and no existing fixture could catch a regression there, since every file in the seek suite has `initial_padding=0`. The new test generates its own Ogg/Opus fixture (`initial_padding=312`, ffmpeg CLI at test time, skipped when unavailable) and asserts a seeked slice matches the decode-and-discard reference within 2e-7 amplitude. With the `pkt_timebase` assignment reverted it fails with "misaligned by -312 samples" — exactly the Opus mandated pre-skip; on this branch it passes.

### DIFF
--- a/src/algorithms/io/audioloader.cpp
+++ b/src/algorithms/io/audioloader.cpp
@@ -45,7 +45,121 @@ void AudioLoader::configure() {
     av_log_set_level(AV_LOG_QUIET);   // choices: {AV_LOG_VERBOSE, AV_LOG_QUIET}
     _computeMD5 = parameter("computeMD5").toBool();
     _selectedStream = parameter("audioStream").toInt();
+
+    _startTime = parameter("startTime").toReal();
+    _endTime   = parameter("endTime").toReal();
+    // Same rule Trimmer has always enforced for the composites that expose these parameters
+    // (EasyLoader, EqloudLoader): an inverted range is an error, an EMPTY range is not -- it
+    // simply yields no audio. Rejecting startTime == endTime here would break callers that
+    // work today.
+    if (_endTime < _startTime) {
+        throw EssentiaException("AudioLoader: 'startTime' cannot be larger than 'endTime'");
+    }
+
     reset();
+}
+
+
+// How much audio to decode-and-discard BEFORE the requested startTime.
+//
+// Seeking a compressed stream lands on a packet boundary, never on a sample, and the packet
+// it lands on is not necessarily decodable in isolation: MP3 carries a bit reservoir across
+// frames and every MDCT codec needs the previous window to reconstruct the current one. So a
+// seek that jumps straight to the target packet produces samples that differ from the ones a
+// decode-from-zero would have produced at the same position -- not by much, but audibly and
+// measurably, and enough to move a BPM or key estimate.
+//
+// The fix is the standard one: ask for a position SEEK_PREROLL_SEC earlier than we want, then
+// decode forward and throw the result away until the exact target sample. By the time the
+// target arrives the decoder has been running long enough for its history to be identical to
+// the from-zero case, and the output is bit-exact (verified in the test suite).
+//
+// 0.5 s is ~19 MP3 frames / ~21 AAC frames, far beyond any codec's actual history, and costs
+// well under a millisecond of decode. Making it bigger buys nothing; making it much smaller
+// starts to matter for codecs with long overlap windows.
+static const Real SEEK_PREROLL_SEC = 0.5;
+
+
+// Position the demuxer at (startTime - preroll) and put the decoder in a clean state.
+//
+// This is the whole of the actual "seek". It is deliberately conservative: AVSEEK_FLAG_BACKWARD
+// guarantees ffmpeg lands at or BEFORE the requested timestamp (never after, which would lose
+// audio the caller asked for), and if the seek fails for any reason we simply do not seek --
+// _startSample trimming in copyFFmpegOutput() then degrades to the old decode-and-discard
+// behaviour, which is slow but still CORRECT. A format with a broken index gets the answer it
+// always got, not an error.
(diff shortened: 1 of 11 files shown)


SimplicityGuy:
While this was generated by Claude, I reviewed the code before issuing the PR. :)

SimplicityGuy:
The failing check (`macos-15-intel / cibuildwheel-tensorflow`) is unrelated to the seeking changes: the macOS CI jobs install an unpinned Homebrew ffmpeg, which now delivers FFmpeg 8/9, and master's `audiocontext.cpp` no longer compiles there (`AVCodec::sample_fmts` was removed in FFmpeg 8) — the seeking code itself compiled fine in the same run. The fix is proposed separately in #1540 (port to `avcodec_get_supported_config()`, preserving the supported-format fallback); this branch temporarily carries the same commit so its CI can run green. When #1540 (or an equivalent) lands on master, a rebase drops the duplicate here automatically (identical patch).

Issue #1516, opened 2026-06-09
ffmpeg7.1 not detected ("ch_layout" build error)


kiefernforst:
Dear devs,

I'm trying to compile git master of essentia against ffmpeg 7.1.4 (most recent of the 7.x series). Sadly it is not detected and the build fails despite https://github.com/MTG/essentia/pull/1494 :

```
[271/306] Compiling src/algorithms/rhythm/beattrackermultifeature.cpp
[...]

Issue #1486, opened 2025-08-16
ModuleNotFoundError: No module named 'essentia.tensorflow' on macOS ARM (osx-arm64) via pip


andrelcbr:
Hello Essentia Team,

I am experiencing an issue where the `essentia.tensorflow` submodule cannot be found on a macOS ARM (Apple Silicon) machine, despite the `essentia-tensorflow` package being successfully installed via pip.

**System Details:**
- **OS:** macOS (ARM / Apple Silicon)
- **Python:** 3.11.13 (installed via Conda/conda-forge)
- **Essentia Version:** 2.1-beta6-dev (installed via pip)
- **Essentia-Tensorflow Version:** 2.1-beta6-dev (installed via pip)

**Problem Description:**
After installing both `essentia` and `essentia-tensorflow` in a clean conda environment, any attempt to `import essentia.tensorflow` results in a `ModuleNotFoundError`.

As a result, TensorFlow-dependent prediction algorithms like `TensorflowPredictEffnetDiscogs` are not available in the `essentia.standard` namespace. A diagnostic script confirmed this.

**Diagnostic Output:**

> Python 3.11.13 | packaged by conda-forge | (main, Jun  4 2025, 14:52:34) [Clang 18.1.8 ]
> Type "copyright", "credits" or "license" for more information.
> 
> IPython 8.37.0 -- An enhanced Interactive Python. Type '?' for help.
> 
> Restarting kernel...
>  
> 
> 
> 
> %runfile '/Users/andrecarvalho/Documents/Músicas/Similaridade/similaridade v3/check essentia.py' --wdir
> --- ESSENTIA MODULE DIAGNOSTICS ---
> Essentia Version: 2.1-beta6-dev
> Essentia Path: /opt/anaconda3/envs/env_similaridade/lib/python3.11/site-packages/essentia/__init__.py
> 
> Contents of 'essentia.standard':
[...]

Issue #1478, opened 2025-07-02
Essentia-Tensorflow on ARM fail compilation


NeptuneHub:
Hi everyone,
I need to compile essentia-tensorflow to run on ARM. More in details I want to create a container with essentia-tensorflow that then can run my open-source app based on essentia AND tensorflow on ARM.

I was ableto create a dockerfile that work on Intel because, basically, I didn't recompile nothing:

```
FROM ubuntu:22.04

ENV LANG=C.UTF-8 \
    PYTHONUNBUFFERED=1 \
    DEBIAN_FRONTEND=noninteractive

WORKDIR /app

RUN apt-get update -o Acquire::Retries=5 -o Acquire::Timeout=30 && \
    apt-get install -y --no-install-recommends \
    python3 python3-pip python3-dev \
    libfftw3-3 libyaml-0-2 libtag1v5 libsamplerate0 \
    ffmpeg wget git vim \
    redis-tools curl \
    supervisor \
    strace \
    procps \
    iputils-ping \
    libopenblas-dev \
    liblapack-dev \
    # Added dependencies for psycopg2-binary
    libpq-dev \
    gcc \
    && rm -rf /var/lib/apt/lists/* \
    && apt-get clean

RUN pip3 install --no-cache-dir numpy==1.26.4

RUN pip3 install --no-cache-dir \
    Flask \
    Flask-Cors \
    redis \
    requests \
    scikit-learn \
    rq \
    pyyaml \
    six \
    # Added psycopg2-binary for PostgreSQL connectivity
    psycopg2-binary \
    ftfy \
    flasgger \
    sqlglot \
    google-generativeai

RUN pip3 install --no-cache-dir essentia-tensorflow

# Create the model directory
RUN mkdir -p /app/model

# Download models from the GitHub release
RUN wget -q -P /app/model \
    https://github.com/NeptuneHub/AudioMuse-AI/releases/download/v1.0.0-model/audioset-vggish-3.pb \
    https://github.com/NeptuneHub/AudioMuse-AI/releases/download/v1.0.0-model/danceability-audioset-vggish-1.pb \
    https://github.com/NeptuneHub/AudioMuse-AI/releases/download/v1.0.0-model/danceability-msd-musicnn-1.pb \
    https://github.com/NeptuneHub/AudioMuse-AI/releases/download/v1.0.0-model/mood_aggressive-audioset-vggish-1.pb \
[...]

Issue #1467, opened 2025-03-20
Successful installation but undefined symbol: avcodec_decode_audio4 when running python


Vriska:
When running the python unit tests via python3 waf run_python_tests, I get this 

python3 waf run_python_tests
/home/vriska/NEW/essentia/wscript:211: SyntaxWarning: invalid escape sequence '\w'
  """
Running python unit tests using /home/vriska/signal/bin/python3
Traceback (most recent call last):
  File "/home/vriska/NEW/essentia/test/src/unittests/all_tests.py", line 27, in <module>
    import essentia
  File "/home/vriska/NEW/essentia/build/python/essentia/__init__.py", line 1, in <module>
    from . import _essentia
ImportError: /home/vriska/NEW/essentia/build/python/essentia/_essentia.cpython-312-x86_64-linux-gnu.so: undefined symbol: avcodec_decode_audio4

This might be relevant, I had some trouble with the installation -

First, I had this https://github.com/MTG/essentia/issues/1411 

Manually installing ffmpeg 4.4 seemed to fix the build issues except for one thing - 

python3 waf gave me this, but this didn't seem to be a problem when i did sudo python3 waf

>[326/348] Compiling src/python/pytypes/vectorstring.cpp
In file included from ../src/python/typedefs.h:27,
                 from ../src/python/pytypes/vectorinteger.cpp:20:
../src/python/pytypes/vectorinteger.cpp: In static member function ‘static PyObject* VectorInteger::toPythonRef(essentia::RogueVector<int>*)’:
../src/python/pytypes/vectorinteger.cpp:30:60: error: ‘PyArray_INT’ was not declared in this scope; did you mean ‘PyArray_Put’?
   30 |   if (dim > 0) result = PyArray_SimpleNewFromData(1, &dim, PyArray_INT, &((*v)[0]));
      |                                                            ^~~~~~~~~~~
../../../signal/lib/python3.12/site-packages/numpy/_core/include/numpy/ndarrayobject.h:122:46: note: in definition of macro ‘PyArray_SimpleNewFromData’
  122 |         PyArray_New(&PyArray_Type, nd, dims, typenum, NULL, \
      |                                              ^~~~~~~
../src/python/pytypes/vectorinteger.cpp:31:52: error: ‘PyArray_INT’ was not declared in this scope; did you mean ‘PyArray_Put’?
   31 |   else         result = PyArray_SimpleNew(1, &dim, PyArray_INT);
      |                                                    ^~~~~~~~~~~
[...]

PR #1542, opened 2026-09-03
Ship linux/aarch64 and floor-correct macOS wheels, vendor libtensorflow from pinned sources, and fix the source-build fallback

### LINKED ISSUE #1541: essentia 2.1b6.dev1438 x86_64 wheels fail to import on Debian slim images: libatomic.so.1 not found (spurious -latomic from the builder image's FFmpeg .pc files)
### Summary

The current `essentia` release on PyPI, `2.1b6.dev1438` (the only release with cp314 wheels), cannot be imported on Debian/Ubuntu "slim" images or any system without `libatomic1` installed: the x86_64 manylinux wheels' `_essentia*.so` declares a `NEEDED libatomic.so.1` that the previous release did not have.

```
$ docker run --rm python:3.14-slim-bookworm bash -c \
    'pip install essentia==2.1b6.dev1438 && python -c "import essentia"'
...
ImportError: libatomic.so.1: cannot open shared object file: No such file or directory
```

### Bisect

| wheel | `NEEDED` on `_essentia*.so` | vendored in `essentia.libs/` |
| --- | --- | --- |
| `essentia-2.1b6.dev1389-cp312-cp312-manylinux_2_17_x86_64` | libstdc++, libm, libgcc_s, libpthread, libc, ld-linux | (none) |
| `essentia-2.1b6.dev1438-cp314-cp314-manylinux2014_x86_64` | **libatomic.so.1**, libdrm-827b956f.so.2.4.0, libstdc++, libm, libgcc_s, libpthread, libc, ld-linux | libdrm-827b956f.so.2.4.0 |

`libatomic.so.1` is on the manylinux allowlist, so `auditwheel repair` leaves it as an external dependency instead of vendoring it, and Debian's `python:*-slim` images (and many minimal containers) do not ship `libatomic1`.

### Cause

The dependency is spurious. `nm -D --undefined-only` on the extension finds no `__atomic_*` / `libat_*` symbols; it uses no atomics. The `-latomic` comes from the FFmpeg 7.1.1 `.pc` files in the `mtgupf/essentia-builds:manylinux2014_x86_64` builder image (`libavutil.pc` and friends record `-latomic` in `Libs`), and the toolchain links with `--no-as-needed`, so `ld` records the `NEEDED` whether or not a symbol is used. The 1389 → 1438 change of builder image (FFmpeg 7.1.1) is what introduced it, and the same image update added the `libdrm` dependency that auditwheel now vendors.

It went unnoticed because the x86_64 wheel's only post-build check is cibuildwheel's `test-command`, which runs inside the manylinux container where `libatomic` exists.

### Fix

Strip the flag before configure in the Linux `before-all` (it cannot be satisfied statically: devtoolset-10's `libatomic.a` is not PIC and fails with `relocation R_X86_64_32 against hidden symbol 'libat_load_16_i1'` when linked into a shared object):

```
sed -i -e 's/ -latomic//g' /usr/local/lib/pkgconfig/lib*.pc
```

and add a clean-container check that installs the built wheel into `python:3.12-slim` / `python:3.14-slim-bookworm` and decodes audio, so this class of break fails CI instead of reaching PyPI. I have this as a self-contained commit (touches `cibuildwheel.toml` and the wheels workflow only) and will include it in a PR shortly; happy to split it out into its own PR if you prefer to land it first.

Workaround for users until a fixed release: `apt-get install -y libatomic1` in the image, or pin `essentia==2.1b6.dev1389` where a cp39–cp313 wheel is acceptable.

SimplicityGuy:
Ship installable `essentia` and `essentia-tensorflow` wheels for every platform phaze and most users actually run on, and make the source-build fallback fail early and clearly instead of deep in a compile.

> **Stacked on #1540** (FFmpeg 8/9 `avcodec_get_supported_config()` port, macOS arm64 cp313 arch-flag fix, numpy<2.5 manylinux constraint). Its three commits are included here because the macOS jobs cannot build against Homebrew FFmpeg 8/9 without them; merge #1540 first and they drop out of this diff, or take them as part of this PR. The remaining 26 commits are cherry-picks of the same commits reviewed and CI-tested on my fork (SimplicityGuy/essentia, branch `wt/bead/epic/ess-rirg`), where the full matrix incl. the aarch64 runners ran green.
>
> **These issues close on the next release that ships the wheels, not at merge time**: published artifacts cannot be fixed retroactively. Related: #1516, #1308, #1350, #1467, #1478, #1486, #1462, #1454, #1446. Fixes #1541 directly.

## What ships from the wheels workflow after this

| platform | `essentia` | `essentia-tensorflow` | libtensorflow source |
| --- | --- | --- | --- |
| linux x86_64 (manylinux2014) | cp39–cp314 | cp39–cp314, ~136 MB | Google tarball 2.18.1 (glibc 2.17) |
| **linux aarch64 (manylinux_2_28)** | **cp39–cp314, new** | **cp39–cp314, ~135 MB, new** | Homebrew 2.21.0 bottle, `arm64_linux` rebuild 1 (glibc 2.27) |
| macOS arm64 (`macosx_15_0`) | cp39–cp314 | cp39–cp314, ~133 MB | Homebrew 2.21.0 bottle, `arm64_sequoia` rebuild 1 (min OS 15.0) |
| macOS x86_64 (`macosx_15_0`) | cp39–cp314 | cp39–cp314, ~154 MB | Google tarball 2.16.2 (min OS 10.15) |
| sdist | yes, new | | |

The user contract is unchanged: `essentia-tensorflow` still vendors the TensorFlow C library and declares no dependency on the `tensorflow` pip package, so it works on every CPython including 3.14, which TensorFlow itself has no wheels for. What changes is where that library comes from: one script, `packaging/fetch_libtensorflow.sh`, with every URL, version and SHA256 pinned in `packaging/build_config.sh`. This removes the `MACOSX_DEPLOYMENT_TARGET=26.2` coupling to whatever `libtensorflow` bottle `brew install` happened to pour (the `arm64_sequoia` build declared minos 26.2, Homebrew/homebrew-core#302034, fixed in Homebrew/homebrew-core#302294 on 2026-09-07); the macOS wheels are tagged `macosx_15_0` again, set by the ordinary audio bottles. On the two arm64 platforms the library is Homebrew's rebuilt 2.21.0 bottle, fetched by digest from ghcr.io so a future rebuild cannot move the pin; on the two x86_64 platforms it is Google's tarball, because the Linux bottle needs glibc 2.27 and the manylinux2014 image has 2.17, and because Homebrew no longer publishes an Intel macOS bottle at all.

## Changes, by area

**Configure / source build**
- FFmpeg floor raised to what the code needs (`AVChannelLayout`, FFmpeg ≥ 5.1; libavcodec ≥ 59.37) with an early, precise failure naming the found version and `.pc` path, plus a compile probe. The old check accepted LibAv 10 from 2015 and let old FFmpeg fail mid-compile on `ch_layout` (#1516).
- `--with-tensorflow` now link-tests the C API (`TF_Version`, `TF_DeleteSession`) at configure time with a 2.x floor and actionable diagnostics, instead of trusting that a `tensorflow.pc` exists and failing at `import` with `undefined symbol: TF_DeleteSession` (#1350, #1308). The include path from Homebrew/`generate-pc.sh` style `.pc` files is normalised. A 2015 vendored copy of `tensorflow/c/c_api.h` that shadowed every installed header is removed.
- `src/3rdparty/tensorflow/setup_tensorflow.py` rewritten for TensorFlow ≥ 2.13 pip wheels (`libtensorflow_cc` layout); the TF 1.x wrapper scripts, which could not work against any installable release, are removed.
- Opt-in `--with-tensorflow-pip-rpath` / `ESSENTIA_TENSORFLOW_PIP_RPATH` for builds that want to resolve TensorFlow from a pip `tensorflow` package at load time (off by default; documented).

**Packaging / CI**
- Linux aarch64 wheels on native `ubuntu-24.04-arm` runners (no QEMU), static 3rdparty deps built in-container behind a cache (four aarch64 fixes to the dependency scripts; gaia is not built on aarch64: it needs Qt 4.8, which has no AArch64 support).
- `essentia-tensorflow` for aarch64 on the same job; libtensorflow fetched from a pinned source on every platform; macOS x86_64 tensorflow job's `test-command` re-enabled.
- The spurious `libatomic.so.1` dependency stripped from the x86_64 wheels (the builder image's FFmpeg `.pc` files carry `-latomic`; the extension uses no atomics). The published `2.1b6.dev1438` x86_64 wheel fails to import on Debian slim images because of it (#1541).
- An sdist job plus from-source install verification on Python 3.12 and 3.14 against apt dependencies (#1462; also adds the missing `essentia.pc.in` to the sdist and invokes the 3rdparty script with bash explicitly).
- `test_wheels`: the Python unit test suite runs against the *installed* cp312 wheel on Linux x86_64, Linux aarch64 and macOS arm64, plus a TensorflowPredictEffnetDiscogs smoke where a tensorflow wheel exists. First time the suite has run against a packaged wheel; it fixed Python 3.12/numpy breakage in the tests, one binding bug (lists of `numpy.float32` rejected by `convertData`, related to #1098), and documents 43 per-method skips with measured deltas in `test/src/unittests/ci_skiplist.txt` for follow-up.
- Every `essentia-tensorflow` wheel runs real inference (effnet, `(N, 1280)`) in cibuildwheel's `test-command` for all six interpreters, and asserts it vendors libtensorflow and declares no `tensorflow` dependency; every wheel gets a clean-container/clean-venv import-and-decode check.

**Docs**
- `installing.rst`: FFmpeg floor and diagnostics, providing the TensorFlow C API, building a wheel against a pip TensorFlow, installing from the sdist. `docs/spikes/ess-rirg.4-tf-c-library-sourcing.md` records the evidence behind the libtensorflow sourcing decision.

## CI

Full matrix on this exact branch (fork CI, same workflow): https://github.com/SimplicityGuy/essentia/actions/runs/33779767704 (14/14: six wheel jobs incl. both aarch64 rows, sdist, from-source installs on 3.12 and 3.14, unit suite against installed wheels on Linux x86_64, Linux aarch64 and macOS arm64). The two docs-only commits on top re-ran as https://github.com/SimplicityGuy/essentia/actions/runs/33784090499 (14/14).

The follow-up commit that moves macOS arm64 to the Homebrew bottle (below) ran green on this repo's own checks (15/15 on b0471014, https://github.com/MTG/essentia/actions/runs/34165817269) and on the fork (https://github.com/SimplicityGuy/essentia/actions/runs/34165813361).

## Notes for reviewers
- `TENSORFLOW_VERSION` (the Bazel build in `packaging/debian_3rdparty/build_tensorflow.sh`) is unchanged at 2.17.0; the four per-platform prebuilt pins are independent of it.
- The vendored libtensorflow versions differ per platform (2.16.2 / 2.18.1 / 2.21.0) because no single release exists for all four slots; all export the 29 `TF_*` functions essentia uses. Both arm64 slots are on the Homebrew 2.21.0 bottle now that Homebrew/homebrew-core#302294 fixed the deployment target; the two x86_64 slots cannot follow (glibc 2.17 on the manylinux2014 image, and no Intel macOS bottle exists any more).
- CI-only skips in the test suite are per-method and reasoned; none touch the io/AudioLoader, streaming or FrameCutter suites.

## Update 2026-09-07: macOS arm64 now vendors the Homebrew bottle

Homebrew/homebrew-core#302294 merged and the `libtensorflow` 2.21.0 bottles were rebuilt. Measured from `LC_BUILD_VERSION` after pulling the rebuild-1 blobs from ghcr.io: `arm64_sequoia` reports minos 15.0 on all nine Mach-O files (was 26.2), `arm64_linux` still tops out at GLIBC_2.27, and the Intel `sonoma` bottle is gone from the formula. So the last commit (`packaging: vendor Homebrew's libtensorflow bottle on macOS arm64`):

- Pins macOS arm64 to the `arm64_sequoia` rebuild-1 digest (TensorFlow 2.21.0 instead of the frozen Google 2.18.1 channel) and bumps linux/aarch64 to the `arm64_linux` rebuild-1 digest. macOS x86_64 and Linux x86_64 keep their Google tarballs, for the reasons in the table.
- Teaches `packaging/fetch_libtensorflow.sh` what a macOS bottle needs that a tarball did not. The bottle is `cellar: :any`, so its dylib ids are `@@HOMEBREW_PREFIX@@/opt/libtensorflow/lib/<name>` placeholders that `brew` rewrites at pour time; linked as-is, delocate cannot resolve them. The script rewrites each id to `<prefix>/lib/<name>` with `install_name_tool -id` and re-signs ad hoc (the arm64 kernel kills a process that loads an invalidly signed dylib: exit 137 before the re-sign, `TF_Version: 2.21.0` after), then asserts no placeholder remains in any load command. It also stops copying the bottle's own read-only `tensorflow.pc`, which made the script's own `.pc` write fail with `Permission denied` for any non-root user; the Linux aarch64 job never hit this because the build container runs as root.
- Rewrites the rationale comments, the `installing.rst` platform table and sourcing paragraph, and adds a dated postscript to the ess-rirg.4 spike, since each argued from the 26.2 bottle that no longer exists.

Verified locally on macOS 26.6 arm64 (fresh install, rerun onto the same prefix, `codesign -v`, a C program built purely from the generated `tensorflow.pc` runs) and in CI: the `macos-15 / cibuildwheel-tensorflow` log shows both dylibs relocated, `minimum macOS: 15.0`, configure's C API link test passing, six `macosx_15_0_arm64` wheels at 132.5 MB each with `libtensorflow.2.21.0.dylib` vendored by delocate, and `embeddings (9, 1280) float32` from the effnet inference on every interpreter plus the two clean-venv checks.


SimplicityGuy:
Excellent, upstream fix to homebrew has been merged! (See linked ticket.) I'm going to update this PR based on that.

